PR #1, opened 2014-10-16
Djnro0.9

simonrogergray:
branch of REANNZ/master fulled updated from upstream grnet:master and with the non-REANNZ specific changes added


kroustou:
Sorry for the delay, we did not see the pull request on time. 
Anyhow we merged it now :)

PR #6, opened 2015-11-10
A few fixes for parse_institution_xml

vladimir-mencl-eresearch:
- Fix: pass write reference between methods
- Verbosity: change a print to write
- Verbosity: always print ERROR messages


zmousm:
Thanks @vladimir-mencl-eresearch !

PR #9, opened 2016-02-02
Allow AltLogin to redirect to target URLs

vladimir-mencl-eresearch:
Allow using `/altlogin` (local user accounts) with a ``?next` parameter.

This aims to support deployments where no SAML federation exists yet and the NRO would prefer not to depend on social login.  With this change, it is possible to use `/altlogin?next=/manage` as a login URL for Institutional Administrators.

Implementation-wise, in `djnro/templates/overview/login.html`, when a "next" parameter exists in the context, reuse it for the redirect - instead of forcing redirect to `/overview`.

### DIFF
--- a/djnro/templates/overview/login.html
+++ b/djnro/templates/overview/login.html
@@ -37,7 +37,11 @@
 							<div class="form-group">
 								<div class="controls">
 									<button type="submit" class="btn">Sign in</button>
+                                                                        {% if next %}
+									<input type="hidden" name="next" value="{{ next }}" />
+                                                                        {% else %}
 									<input type="hidden" name="next" value="{% url overview %}" />
+                                                                        {% endif %}
 								</div>
 							</div>
 						</form>


zmousm:
Good to merge, however wouldn't you also need other changes in order to use this login form? For example `manage_login()` in `edumanage/views.py`? I'm just trying to get a better grasp, as switching to local login wasn't so simple IIRC.

vladimir-mencl-eresearch:
Hi Zenon,

Well, so far, the local login has to be accessed by going to /altlogin directly - the link is not shown on the login page yet.

This PR was to open the door to using it (and check that you don't see an issue with removing the forced redirect to /overview).

I'm thinking about adding a configuration option to list the allowed login methods, and then modifying the login page to show buttons just for the methods configured.

And I was planning that for a another PR to come later - with this one just making the login possible - with an explicit URL provided to the user, pointing at `/altlogin?next=/manage`.

Please let me know if you're fine with this plan.

Cheers,
Vlad

zmousm:
Yes that sounds fine. I understand such options would affect what buttons are displayed through the `djnro/templates/edumanage/welcome_manage.html` template.

vladimir-mencl-eresearch:
Yup, that's the plan - make the template iterate over a list of login methods. Thanks for merging this one!

PR #7, opened 2016-01-28
Minor fix: Update links from Institution to InstServer

vladimir-mencl-eresearch:
Update code linking from Institution to InstServer to reflect changes in the
model introduced in 15691d3 - instead of instserver_set, use the relation name
"servers" now explicitly given in the model.

PR #10, opened 2016-02-15
Minor fixes from #7

vladimir-mencl-eresearch:
Hi Zenon,

I have cherry-picked the two commits from #7 that are not related to the InstServer M2M change - please let me know if you are happy to merge these two minor fixes.

Cheers,
Vlad


zmousm:
Thanks for the template fix.

As for `requirements.txt`:
- Indeed PyYAML is missing; but in order to track the versions of other dependencies (still mostly aligned to Debian Wheezy), I would write the version argument as `>=3.0.10`.
- Mako is only necessary for (example) templates under extras. As that actually won't typically run in the same system/environment as DjNRO, I would be inclined not to add it in `requirements.txt` or perhaps to add it in `optional_requirements.txt`. What do you think?

Also, please rebase this PR on the rewritten master branch.

vladimir-mencl-eresearch:
Hi Zenon, 

Most packages are listed as pinned to exact versions - but I've updated the PR as requested and rebased - can you please merge?

Cheers,
Vlad

PR #12, opened 2016-02-18
Two fixes related to the Instserver M2M changes (reinstate #7)

vladimir-mencl-eresearch:
Hi Zenon,

These are the fixes I originally pushed in #7 - now in a separate branch.

Cheers,
Vlad

### DIFF
--- a/edumanage/forms.py
+++ b/edumanage/forms.py
@@ -65,22 +65,23 @@ class InstServerForm(forms.ModelForm):
 
     def clean_ertype(self):
         ertype = self.cleaned_data['ertype']
-        institution = self.cleaned_data['instid']
-        inst_type = institution.ertype
-        type_list = [inst_type]
-        if inst_type == 3:
-            type_list = [1, 2, 3]
-        if ertype:
-            if ertype not in type_list:
-                raise forms.ValidationError('Server type cannot be different than institution type (%s)' %dict(self.fields['ertype'].choices)[inst_type])
-            return self.cleaned_data["ertype"]
-        else:
-            raise forms.ValidationError('This field is required.')
+	if not ertype:
+	    raise forms.ValidationError('This field is required.')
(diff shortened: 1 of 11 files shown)


edumanage/forms.py
```diff
@@ -97,8 +100,10 @@ def clean_acct_port(self):
 
     def clean_rad_pkt_type(self):
         rad_pkt_type = self.cleaned_data['rad_pkt_type']
-        institution = self.cleaned_data['instid']
-        if institution.ertype in [1,3]:
+	if not 'ertype' in self.cleaned_data:
+                raise forms.ValidationError(_('The Type field is required to validate this field.'))
+        ertype = self.cleaned_data['ertype']
+        if ertype in [1,3]:
```
zmousm, line 106: This field is actually required even if the institution/server is SP only, because NRO servers may need to know whether they should handle accounting packets. We want to deprecate accounting and thus get rid of RADTYPES in the future, but this is how things stand for now. So I believe you should remove any 'ertype' checking in this function.

zmousm:
Hi Vladimir. Thanks for this. Apart from the line note, LGTM!

Please rebase on branch next:
- next is rewritten master branch with the instserver_m2m commit
- master is the rewritten branch **without** the instserver_m2m commit

edumanage/forms.py
zmousm, line 106: Sorry, scratch this comment, it is not true: there is no use of `rad_pkt_type` for SP only servers/institutions. Your approach is fine and it is more appropriate than checking institution `ertype`. Eventually we should also remove/hide this field from the form for such cases.

zmousm:
Rebase on next and it's ready to merge.

vladimir-mencl-eresearch:
Re-submitted as #13 with next as the target for merging.

PR #14, opened 2016-02-24
InstServer M2M: fix instid handling in server edit

vladimir-mencl-eresearch:
Fix issue: editing and saving an InstServer by an inst admin drops all other Inst associations

Fix: remove field['instid'] overrides

Subsequent issue: this allows arbitrary edits of the list of institutions IF
the user changes the display:none style in browser.

Fix: exclude ['instid'] from the form

Subsequent issue: this breaks validation relying on instid

Attempted fix: access instid from model instance, not from cleaned_data: self.instance.instid.all()
Reason for not using this fix: breaks on creating a new server:

```
ValueError: "<InstServer: Server: , Type: None>" needs to have a value for field "instserver" before this many-to-many relationship can be used.
```

Alternate fix: hint the list of institutions to the ModelForm out-of-band (via inst_list attribute)

Additional fix: add the institution to the InstServer AFTER saving.

### DIFF
--- a/edumanage/forms.py
+++ b/edumanage/forms.py
@@ -62,12 +62,13 @@ class InstServerForm(forms.ModelForm):
 
     class Meta:
         model = InstServer
+        exclude = ['instid']
 
     def clean_ertype(self):
         ertype = self.cleaned_data['ertype']
 	if not ertype:
 	    raise forms.ValidationError('This field is required.')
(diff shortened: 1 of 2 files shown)


zmousm:
Impressive pivoting through this problem :+1: Thanks!

PR #17, opened 2016-04-04
 Branding: make NRO inst name configurable 

vladimir-mencl-eresearch:
There had been a few places where a message displayed to end users would have the name GRNET hard-coded.

Make this configurable - add new setting

NRO_INST_NAME (defaulting to "GRNET")

and also

NRO_FEDERATION_NAME (defaulting to "GRNET AAI federation") to allow having a separate name for the federation.

### DIFF
--- a/djnro/local_settings.py.dist
+++ b/djnro/local_settings.py.dist
@@ -119,6 +119,8 @@ NRO_COUNTRY_NAME = _('My Country')
 NRO_COUNTRY_CODE = 'tld'
 # main domain url used in right top icon, eg. http://www.grnet.gr
 NRO_DOMAIN_MAIN_URL = "http://www.example.com"
+# NRO federation name
+NRO_FEDERATION_NAME = "GRNET AAI federation"
 # developer info for footer
 NRO_PROV_BY_DICT = {"name": "GRNET NOC", "url": "//noc.grnet.gr"}
 #provider social media contact (Use: // to preserve https)
--- a/djnro/local_settings.py.dist
+++ b/djnro/local_settings.py.dist
@@ -113,9 +113,7 @@ TIME_ZONE = 'Europe/Athens'
 # map center (lat, lng)
 MAP_CENTER = (36.97, 23.71)
 
-# Frontend country specific vars, eg. Greece
-NRO_COUNTRY_NAME = _('My Country')
-# Variable used by context_processor to display the "eduroam | <country_code>" in base.html
+# Variable used to determine the active Realm object (in views and context processor)
 NRO_COUNTRY_CODE = 'tld'
 # main domain url used in right top icon, eg. http://www.grnet.gr
 NRO_DOMAIN_MAIN_URL = "http://www.example.com"
--- a/djnro/templates/front/management.html
+++ b/djnro/templates/front/management.html
@@ -17,7 +17,7 @@
     <h5>AAI Federation</h5>
     <hr>
     <p>{% blocktrans %}Authentication and authorization are carried out through a{% endblocktrans %} <a href="http://shibboleth.net/products/service-provider.html">Shibboleth SP</a>.</p>
-    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} <a href="http://aai.grnet.gr/documentation">{% trans "policy and procedures documentation" %}</a> {% trans "provided by the GRNET AAI federation" %}:
+    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} <a href="http://aai.grnet.gr/documentation">{% trans "policy and procedures documentation" %}</a> {% trans "provided by the" %} {{ FEDERATION_NAME }}:
         <table class="table table-bordered">
             <thead><tr><th>{% trans "Attribute" %}</th><th>{% trans "Description" %}</th></tr></thead>
             <tbody>
--- a/edumanage/context_processors.py
+++ b/edumanage/context_processors.py
@@ -1,11 +1,13 @@
 from django.conf import settings
+from edumanage.models import Realm
 
 
 def country_code(context):
     # return the value you want as a dictionnary. you may add multiple values in there.
     return {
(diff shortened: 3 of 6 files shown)


vladimir-mencl-eresearch:
Hi Zenon,

Wondering what you think about this change.  I've now realized that the model possibly already contains the NRO Institution name (Realm.org_name via Name_i18n), so that could be an alternative - but pulling a Realm model instance (singleton) might make the code too fragile - so I'm still inclined to go with the approach in this PR.

What do you think?

Cheers,
Vlad

zmousm:
Hi Vladimir,

No objections regarding `NRO_FEDERATION_NAME`, in the longer term however I believe this page should move to a flat page or something site-specific, since the variety of authentication and authorization scenarios renders this content unsuitable for reuse.

As for `NRO_INST_NAME`, there is indeed a logical gap here. As you noted, we do have realms (not a singleton), per the eduroam database schema. In a multi-tenant scenario, there could be more than one realm. Institutions are tied to a realm, so the backend (edumanage) could render everything targeting the respective realm. For the front-end a realm could be chosen through the django sites framework or a custom middleware to handle url schemes accommodating multiple realms. We are far from such a scenario, however.

Besides realms, we have a number of NRO_ settings referring to the same or complementary things. I don't like the duplication of data, but these are indeed static settings, for the time being.

Nevertheless I wanted to see how we can get what is available through the realm object, rather than settings, just using `settings.NRO_COUNTRY_CODE` to select the active `Realm` object. I am reluctant whether this is an overkill, also from a performance perspective. What do you think of the following patch:

``` diff
diff --git a/djnro/local_settings.py.dist b/djnro/local_settings.py.dist
index 28e2def..07a354b 100644
--- a/djnro/local_settings.py.dist
+++ b/djnro/local_settings.py.dist
@@ -113,9 +113,7 @@ TIME_ZONE = 'Europe/Athens'
 # map center (lat, lng)
 MAP_CENTER = (36.97, 23.71)

-# Frontend country specific vars, eg. Greece
-NRO_COUNTRY_NAME = _('My Country')
-# Variable used by context_processor to display the "eduroam | <country_code>" in base.html
+# Variable used to determine the active Realm object (in views and context processor)
 NRO_COUNTRY_CODE = 'tld'
 # main domain url used in right top icon, eg. http://www.grnet.gr
 NRO_DOMAIN_MAIN_URL = "http://www.example.com"
diff --git a/edumanage/context_processors.py b/edumanage/context_processors.py
index 0b4c01d..e1f5875 100644
--- a/edumanage/context_processors.py
+++ b/edumanage/context_processors.py
@@ -1,11 +1,13 @@
 from django.conf import settings
+from edumanage.models import Realm


 def country_code(context):
     # return the value you want as a dictionnary. you may add multiple values in there.
     return {
-        'COUNTRY_NAME': settings.NRO_COUNTRY_NAME,
         'COUNTRY_CODE': settings.NRO_COUNTRY_CODE,
+        'COUNTRY_NAME': dict(settings.REALM_COUNTRIES)[settings.NRO_COUNTRY_CODE],
+        'NRO_REALM': Realm.objects.get(country=settings.NRO_COUNTRY_CODE),
         'DOMAIN_MAIN_URL': settings.NRO_DOMAIN_MAIN_URL,
         'DOMAIN_HELPDESK_DICT': settings.NRO_DOMAIN_HELPDESK_DICT,
         'MAP_CENTER': settings.MAP_CENTER,
diff --git a/edumanage/models.py b/edumanage/models.py
index 6b4da39..a350a41 100644
--- a/edumanage/models.py
+++ b/edumanage/models.py
@@ -535,6 +535,17 @@ class Realm(models.Model):
             'country': self.country,
         }

+    def get_name(self, lang=None):
+        name = ', '.join([i.name for i in self.org_name.all()])
+        if not lang:
+            return name
+        else:
+            try:
+                name = self.org_name.get(lang=lang)
+                return name
+            except Exception:
+                return name
+

 # TODO: this represents a *database view* "realm_data", find a better way to write it
 class RealmData(models.Model):
diff --git a/edumanage/views.py b/edumanage/views.py
index c8417ac..4b70538 100644
--- a/edumanage/views.py
+++ b/edumanage/views.py
@@ -1624,7 +1624,9 @@ def user_login(request):
                     " Administrators have been notified and will activate"
                     " this account within the next days. <br>If this account"
                     " has remained inactive for a long time contact your "
-                    "technical coordinator or GRNET Helpdesk") % user.username
+                    "technical coordinator or %s Helpdesk") % (user.username,
+                                                               get_nro_name(request.LANGUAGE_CODE))
+                                                                  
                 return render_to_response(
                     'status.html',
                     {'status': status, 'inactive': True},
@@ -1724,7 +1726,8 @@ def selectinst(request):
                 " Administrators have been notified and will activate "
                 "this account within the next days. <br>If this account"
                 " has remained inactive for a long time contact your technical"
-                " coordinator or GRNET Helpdesk") % userprofile.user.username
+                " coordinator or %s Helpdesk") % (userprofile.user.username,
+                                                  get_nro_name(request.LANGUAGE_CODE))
             return render_to_response(
                 'status.html',
                 {'status': error, 'inactive': True},
@@ -2343,3 +2346,8 @@ def lookupShibAttr(attrmap, requestMeta):
             if len(requestMeta[attr]) > 0:
                 return requestMeta[attr]
     return ''
+
+def get_nro_name(lang):
+    return Realm.objects.\
+        get(country=settings.NRO_COUNTRY_CODE).\
+        get_name(lang=lang)
```

vladimir-mencl-eresearch:
Hi Zenon,

Thanks - I very much like this.

This is the way to go - I was concerned it might make the code too fragile, but this is looking just perfect.

I have just rebased this PR on next and reduced the original commit (0e0e59c) to just including the federation name.

And added in your patch (of course giving you authorship) - hope you are fine with this, wanted to keep it in a single PR.

Would you be happy to merge this?

Cheers,
Vlad

vladimir-mencl-eresearch:
Ah, I just realized your version did not cover the change I also did to the status message done in social_active_required in edumanage/decorators.py

I'm now working on it...

Cheers,
Vlad

vladimir-mencl-eresearch:
Done - ready to merge from my end now.

zmousm:
OK, merging. Maybe `get_nro_name()` should move to some more neutral module, but we'll get to that later.

PR #18, opened 2016-07-15
Add Google API key variable

chrubuntu:
...and use it for initial log into the admin interface.

PR #19, opened 2016-08-11
DjNRO "next" misc fixes

vladimir-mencl-eresearch:
In the parse_institution_xml management command, use settings.NRO_COUNTRY_CODE
instead of hard-coded pk=1.  Otherwise, the management command fails when the
primary key of the Realm object is not one.  (E.g., if the object was deleted
and re-created).

The value settings.NRO_COUNTRY_CODE is already used as a key into the Realm
model in edumanage/views.py in get_nro_name.

See also discussion in #18.

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -77,12 +77,12 @@ Set your cache backend (if you want to use one). For production instances you ca
 
 NRO specific parameters. These affect HTML templates::
 
-	# Frontend country specific vars, eg. Greece
-	NRO_COUNTRY_NAME = _('My Country')
-	# Variable used by context_processor to display the "eduroam | <country_code>" in base.html
+	# Variable used to determine the active Realm object (in views and context processor)
 	NRO_COUNTRY_CODE = 'gr'
 	# main domain url used in right top icon, eg. http://www.grnet.gr
 	NRO_DOMAIN_MAIN_URL = "http://www.example.com"
+	# NRO federation name
+	NRO_FEDERATION_NAME = "GRNET AAI federation"
 	# provider info for footer
 	NRO_PROV_BY_DICT = {"name": "EXAMPLE DEV TEAM", "url": "http://devteam.example.com"}
 	#NRO social media contact (Use: // to preserve https)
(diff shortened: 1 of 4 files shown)


vladimir-mencl-eresearch:
Hi @zmousm 

After the discussion in #18 (and on the mailing list), I think it is best if we revert part of f7e106c and remove `NRO_REALM` from edumanage/context_processors.py.  It is not used anywhere and is a trap for those doing a new install.

And I have also updated the documentation - primarily covering the changes I did to `local_settings.py.dist` which never made it into the documentation.

Please let me know whether you'd be happy to merge this.

Cheers,
Vlad

vladimir-mencl-eresearch:
... and I've also added a note clarifying it's important to have NRO_COUNTRY_CODE match both REALM_COUNTRIES and the Realm object (still used in get_nro_name)

vladimir-mencl-eresearch:
Hi Zenon, 

Welcome back from your vacations!

I have found a solution to an issue reported by Johann - where locations were showing up on the map with the name displayed as 'unknown'.

This appears to have been accidentally introduced in 7f1f9f3.

I've added a fix to this PR (and changed the name to "misc fixes" - as that's what the PR now has).

All of the fixes are minor - please let me know whether you'd be happy to merge this.

Cheers,
Vlad

zmousm:
Hi Vladimir.

Terribly sorry for the delay again.

I had already fixed the indentation mess in `get_all_services()` view on our production site, but unfortunately I had put off committing/pushing this.
I took the liberty to push your fix as-is under master, which is where it all began and it should be getting the fix as well.

I don't remember why NRO_REALM was added in the context processor but of course I wouldn't object removing it for now.

Please remove b3f032f and I am ready to merge the rest to "next".

Thanks again.

zmousm:
Please also remove this line from `edumanage/context_processors.py`:

``` python
from edumanage.models import Realm
```

vladimir-mencl-eresearch:
Hi Zenon,

I've created a new PR with the updated commits as #27 - so closing this one.

Cheers,
Vlad

PR #20, opened 2016-08-30
Google API keys

vladimir-mencl-eresearch:
Hi Zenon,

I have taken #18 submitted by Chris and ported that to the `next` branch.

I've made the GOOGLE_API_KEY setting optional to preserve compatibility with existing deployments.

And I've also fixed a few formatting issues that I spotted along the way - improperly encoded '&' in URLs etc.

Please let me know whether happy to merge.

Cheers,
Vlad

### DIFF
--- a/djnro/local_settings.py.dist
+++ b/djnro/local_settings.py.dist
@@ -21,6 +21,9 @@ ALLOWED_HOSTS = ['*']
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = '<put something really random here, eg. %$#%@#$^2312351345#$%3452345@#$%@#$234#@$hhzdavfsdcFDGVFSDGhn>'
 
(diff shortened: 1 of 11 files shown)


zmousm:
Thanks Vladimir.

I believe "API_KEY" as defined in the context processor is too generic for a name. So I would like to change every attribute name/dict key to GOOGLE_MAPS_API_KEY.

This issue is not tied to anything introduced in the "next" branch and it also affects master. So I would like to re-target master.

I don't want to put you in the hassle of re-doing the PR, so I have created #25 with all these commits (@chrubuntu and yours) plus the rename. Please check if that is OK.

vladimir-mencl-eresearch:
Hi Zenon,

I'm happy with how you are handling this in #25, so assuming this would get merged from master to next, I'm closing this PR now.

Cheers,
Vlad

PR #22, opened 2016-09-08
Work-around Django creating spurious migrations for fields with choices pointing to settings ("dynamic")

zmousm:
See
- discussion: https://code.djangoproject.com/ticket/23581
- solution (workaround) employed by django-cms:
  https://github.com/divio/django-cms/issues/3479


zmousm:
@vladimir-mencl-eresearch check this out before I merge

vladimir-mencl-eresearch:
Hi Zenon,

Thanks for the fix!

I was  aware of the issues, and I also realized I accidentally created the Django 1.7+ migrations with New Zealand specific settings - but this solution (making migrations pick values from settings) provides the perfect fix, thanks!

Cheers,
Vlad

PR #23, opened 2016-09-08
Fix requirements.txt (again)

zmousm:
- No need to bump PyYAML min. version to 3.11 (Debian Wheezy provides 3.10
  and that should still work).
- The less than next "release" notation for max. version has the unfortunate
  side effect that a pre-release version apparently passes this spec. So fix
  the spec for Django to ensure that pip will not install anything later than
  1.8.x.

```
  >>> from pkg_resources import parse_version
  >>> parse_version('1.9c1') < parse_version('1.9')
  True
```

  While:

```
  >>> parse_version('1.9c1') < parse_version('1.9a')
  False
```

  See:
  http://stackoverflow.com/questions/11887762/compare-version-strings/21065570#21065570
  https://setuptools.readthedocs.io/en/latest/setuptools.html#specifying-your-project-s-version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django>=1.8.11,<1.9
+Django>=1.8.11,<1.9a
 argparse>=1.2.1
 django-registration==2.0.4
 django-tinymce==1.5.3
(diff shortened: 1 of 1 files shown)


zmousm:
@vladimir-mencl-eresearch fyi

vladimir-mencl-eresearch:
Thanks - I did not run into any issues with a 1.9a version slipping in, but it's looking good to me.

Cheers,
Vlad

PR #24, opened 2016-09-08
Fix branding

### LINKED ISSUE #21: Branding footer does not get initialized
Hi Zenon,

While I was working on #20 yesterday, I noticed that `djnro/templates/partial/footer.html` uses BRANDING context variable, which is however not defined anywhere.  And in turn, the footer looks rather incomplete with unpopulated values.

I've had a look and all the different parts (helpdesk, social_media, service_provided_by) are defined in local_settings.py.dist - and I get most correctly populated by adding `BRANDING` to `edumanage/context_processors.py`:

```
        'BRANDING': {
             'helpdesk': settings.NRO_DOMAIN_HELPDESK_DICT,
             'social_media': settings.NRO_PROV_SOCIAL_MEDIA_CONTACT,
             'service_provided_by': settings.NRO_PROV_BY_DICT,
        },
```

There is a little bit of additional polishing required:
- Drop `DOMAIN_HELPDESK_DICT` from the context parameters (this is not used anywhere)
- Change how the social_media `icon` item is interpreted: sample definition in `local_settings.py.dist` uses paths to images, while the code in `djnro/templates/partial/footer.html` interprets this as a CSS class.

Would you like me to add this to the misc fixes PR #19 ?

Cheers,
Vlad

zmousm:
- Merge relevant keys in a BRANDING (sub)dict
- Don't rename the context processor more appropriately _yet_
- Allow styling similar to MANAGE_LOGIN_METHODS (local_image, image_url
  or fa_style, but keep the older option icon for backwards compatibility)

Fixes #21

### DIFF
--- a/djnro/local_settings.py.dist
+++ b/djnro/local_settings.py.dist
@@ -122,8 +122,8 @@ NRO_DOMAIN_MAIN_URL = "http://www.example.com"
 NRO_PROV_BY_DICT = {"name": "EXAMPLE NRO TEAM", "url": "http://noc.example.com"}
 # social media contact (Use: // to preserve https)
 NRO_PROV_SOCIAL_MEDIA_CONTACT = [
-    {"url": "//facebook.com/example.com", "icon":"/static/img/facebook_img.png", "name":"Facebook"},
-    {"url": "//twitter.com/example_com", "icon":"/static/img/twitter_img.png", "name":"Twitter"},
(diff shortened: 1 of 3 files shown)


vladimir-mencl-eresearch:
Hi Zenon,

Thanks, this is looking great - I assume this would get merged into next too once we are all done.

Cheers,
Vlad

PR #25, opened 2016-09-08
Introduce optional API key for Google Maps

zmousm:
As in #18 and #20 but
- target master
- rename options to GOOGLE_MAPS_API_KEY


chrubuntu:
Looks good to me.

vladimir-mencl-eresearch:
Hi Zenon,

Thanks, this looks all good to me - and I assume this would get merged into `next` once we have all of the currently pending changes resolved.

Cheers,
Vlad

PR #28, opened 2016-09-09
AllServicePoints: fix MultipleObjectsReturned error

vladimir-mencl-eresearch:
Hi Zenon,

I had received an error report from Johann, which translated to my fix that you ported to `master` as 3a60a4c breaking when with unhandled exception MultipleObjectsReturned
when there are multiple names (for either Institution or Service Location) with
the selected language.

And, this is permitted in the model, and we have users who create such DB state.

Fix this by using filter(lang=...) instead of get(lang=...)

And wrap the handling this for both org_name and loc_name in a single generic
function - taking a
django.contrib.contenttypes.fields.GenericRelatedObjectManager reference,
language, default_language and default_value.

This also removes the exception handling clutter from get_all_services().

Cheers,
Vlad

### DIFF
--- a/edumanage/views.py
+++ b/edumanage/views.py
@@ -1525,22 +1525,8 @@ def get_all_services(request):
         else:
             response_location['enc'] = u"-"
         response_location['AP_no'] = u"%s" % (sl.AP_no)
-        try:
-            response_location['inst'] = sl.institutionid.org_name.get(
-                lang=lang
-            ).name
-        except Name_i18n.DoesNotExist:
-            try:
-                response_location['inst'] = sl.institutionid.org_name.get(lang='en').name
-            except Name_i18n.DoesNotExist:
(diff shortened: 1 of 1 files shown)


zmousm:
Hi Vladimir,

Multiple names in the same language don't make a lot of sense, yet indeed neither the eduroam database XSD nor our schema forbid this, so of course it should not break with unhandled exceptions.

Removing the clutter in `get_all_services()` is a good thing! However I just noticed we also added a `get_nro_name()` in the course of `next`. We do have model methods for similar purposes (but they don't handle concatenating multiple names); I am thinking there could be a different way around this, so as to avoid a _proliferation_ of functions dealing with i18n names in the future.

I will also update the other similar view (`get_service_points()`) to use this function.

Thanks for the fix!

PR #31, opened 2016-09-16
Merge latest fixes in master back into next (external merge)

vladimir-mencl-eresearch:
Hi Zenon,

I've tried another take on merging the latest changes in master back into next - I've done a merge outside github, resolved all conflicts, and this is the big merge commit to merge back into next.

All of the conflict resolution was in the end simple.  I did look up the commits on master and next that made change to the adjacent/slightly overlapping code and there was no "new code" being committed in the merge commit - just cleanly separating the changes where the mechanisms in git could not handle this automatically.

The only exception is a change to the default value NRO_FEDERATION_NAME.   When merging this with your commit changing default settings to NRO-neutral, I changed "GRNET AAI Federation" to "AAI Federation".

Everything else was just copy-editing separating adjacent changes, all trivial.

Please let me know whether you are happy to merge this.

Cheers,
Vlad

PS: And I'll just close #30 to avoid having a duplicate merge request.


zmousm:
Vladimir, thanks for this, especially for explaining resolved conflicts in the merge commit message. I have already done this a number of times, so `git rerere` has noted all the right conflict resolutions :) However I am always merging next into master.

If I merged your PR through GitHub or with `--no-ff`, it would create yet another artificial merge commit; basically you've already done all the work I would be doing while merging.

I deployed a 'next merged into master' version to eduroam.gr yesterday, after which I discovered some further issues. I am inclined to get these resolved before merging, so let's wait on this for a few days.

zmousm:
@vladimir-mencl-eresearch I will now close this. There were some further changes to both master and next, after which I (finally!) merged next into master. Thanks again for your effort.

PR #32, opened 2016-11-20
Refactor edumanage/management/commands/parse_institution_xml.py

zmousm:
- More understandable error messages for some show-stopper exceptions
- Switch from xml.etree to lxml.etree
- Parse whole elements (containers) before creating objects: determine if
  required data (per eduroam db XSD) is missing and allow for optional data to
  be absent
- Split XML parsing into "reusable" methods corresponding to individual models
  (where possible, does not apply to Institution, InstitutionDetails)
- Use Model.objects.get_or_create() and try hard to find previously
  added entities so as to not add them again


zmousm:
Merging with location matching strategy extended to comparing names. Using an option for choosing among different strategies still under discussion; if conclusive, it will be added later.

PR #36, opened 2016-11-30
Housekeeping: static files cleanup

zmousm:
* Remove foreign static files
* Move `static/` to `djnro/static/`
* Default tinymce setting and update docs about `collectstatic`


zmousm:
Once this is merged, we may see a radical change in *language details* as reported on GitHub ;)

zmousm:
@vladimir-mencl-eresearch please let me know what you think

PR #40, opened 2016-12-31
Fix custom User model: username MaxLengthValidator also needs patching

zmousm:
In the context of Django 1.8 migration, longerusername was replaced
with a custom User model. Like the former solution, this also patches
the username field to increase max_length. Alternatively one would
need to add a different field, since field name hiding is not
permitted[1], and have USERNAME_FIELD[2] point to it. But this
involves code duplication which seems unnecessary for such a small
change.

However the patching approach also has issues: It is applied after
CharField is instantiated, which means some things are setup before
max_length is adjusted as intended. MaxLengthValidator is such a case,
so it needs patching too. This is also the approach taken[3] by
longerusername.

This issue is evident in the Django admin interface, where editing any
user field and saving leads to a validation error if the username
exceeds the default max_length (30 characters).

Django documentation[4] suggests that UserCreationForm and UserChangeForm
ModelForms used in UserAdmin must also be overridden, so as to have their
Meta class point to the custom User model. This is again the approach[5]
in longerusername (for older Django versions though, no custom User
model involved). In practice our use case apparently does not require
this (on the contrary actually overriding the forms without adjusting
MaxLengthValidator does not have any effect whatsoever).

[1] https://docs.djangoproject.com/en/1.8/topics/db/models/#field-name-hiding-is-not-permitted
[2] https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#django.contrib.auth.models.CustomUser.USERNAME_FIELD
[3] https://github.com/madssj/django-longer-username-and-email/blob/master/longerusernameandemail/models.py#L23
[4] https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#custom-users-and-the-built-in-auth-forms
[5] https://github.com/madssj/django-longer-username-and-email/blob/master/longerusernameandemail/forms.py

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -1,6 +1,5 @@
 from django.db import models
-from django.contrib.auth.models import AbstractUser
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.models import AbstractUser, validators
 from django.contrib import admin
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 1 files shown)


zmousm:
@vladimir-mencl-eresearch take a look; after almost 2 days of head scratching, I believe this is the least intrusive solution that contributes to your previously accepted approach.

I sense this may not be the only/last such issue, according to the included problem description, but I guess we will have to keep fixing them as we find them.

Happy new year!

vladimir-mencl-eresearch:
Hi Zenon,

Sorry, only got to this now.  Looks all good to me - sorry I missed this in 7acf827

Cheers,
Vlad

PR #41, opened 2017-01-18
Support optional sentry integration

ioparaskev:
Add opt in / out(default)  setting for sentry
Support environmental/hardcoded sentry dsn
Validate sentry dsn is setup if sentry is used
Use default empty dict sentry placeholder for backward settings
compatibility


zmousm:
Thanks for this.

Would you please:
- Add `raven` (?) to `requirements-optional.txt`.
- Mention it in `docs/installation/requirements.md` under Conditional Dependencies.
- Add a short section about the steps to enable this integration in `docs/installation/install.md`, say, after Extra Apps. For example add this setting and/or that env variable.

requirements-optional.txt
```diff
@@ -1 +1,2 @@
 Mako==1.0.3
+raven
```
zmousm: Would it make sense to limit the range of versions for the package, as applicable for Django 1.8?

zmousm:
See one minor inline comment. Apart from that, I'm ready to merge and I will fine-tune some wording in the docs afterwards.

requirements-optional.txt
ioparaskev: I'm not sure if we can set a range, maybe a minimum version would be more preferable?

PR #43, opened 2017-03-11
Fix capitalisation Eduroam -> eduroam

ghalse:
This might have come up before and been deliberately ignored, but just in case it has not.

The canonical form of the eduroam trademark is "eduroam" (all lowercase). There really should be definitive references for this, but see https://www.terena.org/mail-archives/mobility/msg01161.html (and https://twitter.com/ln4711/status/647748545546727424 :-)).

Use of other capitalisation, particularly on an NRO website, leads to confusion:

Wireless SSIDs are case sensitive, so eduroam != Eduroam. However, it would appear that some wireless network operators aren't aware of this, and having incorrect capitalisation on the NRO site perpetuates the idea that it is okay to call your wireless network Eduroam. (See https://student.uj.ac.za/docs/Connecting_to_WiFi_pc.pdf for the local example that triggered this...)

From an end-user perspective, it reinforces the perception that a wireless network called Eduroam is part of eduroam. That's bad for all sorts of reasons, not least of which is that it breaks the user experience. (Following on from the above, staff from that institution were confused when they had to configure a new network whilst roaming.)

So in an effort to reinforce the right(tm) way of capitalising it, we should be consistent in our use :)


ghalse:
Thanks for the most useful software, BTW :-)

djnro/templates/front/world.html
```diff
@@ -8,7 +8,7 @@
 
 
 {% block content %}
-<h1>{% trans "Eduroam Worldwide" %}</h1>
+<h1>{% trans "eduroam Worldwide" %}</h1>
```
zmousm, line 11: Worldwide -> worldwide but I'll fix this myself

zmousm:
Thanks for your kind words and sorry for the late reply.

Yes, the capitalization was mostly an oversight, and should be corrected.

PR #44, opened 2017-03-14
Genericise the Management instructions page

ghalse:
This patch replaces the GRnet-specific parts of the management instructions with templated variables that come from local_settings.py, allowing other NROs with other AAI federations to use the templated page more easily. It is specifically backwards compatible, meaning that in the absence of a documentation URL it'll revert to the GRnet AAI one. However, it may make more sense to replace this with None.

### DIFF
--- a/djnro/templates/front/management.html
+++ b/djnro/templates/front/management.html
@@ -17,12 +17,14 @@
     <h5>AAI Federation</h5>
     <hr>
     <p>{% blocktrans %}Authentication and authorization are carried out through a{% endblocktrans %} <a href="http://shibboleth.net/products/service-provider.html">Shibboleth SP</a>.</p>
-    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} <a href="http://aai.grnet.gr/documentation">{% trans "policy and procedures documentation" %}</a> {% trans "provided by the" %} {{ FEDERATION_NAME }}:
+    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} <a href="{{ SHIB_DOCUMENTATION_URL }}">{% trans "policy and procedures documentation" %}</a> {% trans "provided by the" %} {{ FEDERATION_NAME }}:
         <table class="table table-bordered">
             <thead><tr><th>{% trans "Attribute" %}</th><th>{% trans "Description" %}</th></tr></thead>
             <tbody>
             <tr class="success"><td>eduPersonPrincipalName</td><td>{% trans "Provides a string that uniquely identifies an administrator in the management application." %}</td></tr>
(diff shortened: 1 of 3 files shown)


zmousm:
Thanks for improving this; I've been meaning to make it more generic for a long time...

I think FEDERATION_DOC_URL is more appropriate than SHIB_DOCUMENTATION_URL; and yes it makes no sense to revert to the current hard-coded link for GRNET AAI, the template should just render no anchor if the setting is not configured. Please push these changes if possible.

As for SHIB_AUTH_ENTITLEMENT, I agree for making it optional but let me just note this was meant as a SAML-asserted authorization prior to admins' registration; unfortunately nothing equivalent was available (or at least not implemented) for `python-social-auth` providers. More recently however we've been thinking about eventually moving to a registration-after-invitation workflow...

ghalse:
Done.

I actually like the way SHIB_AUTH_ENTITLEMENT is supposed to work, but in our federation, we're not there yet. So for me, the use case for making it optional was to allow federated login but to still use the existing social auth linking. This works if you sent SHIB_AUTH_ENTITLEMENT to a blank, but the documentation was still misleading for end users.

Registration after invitation (like CAT) would cut down on the number of spurious social registrations that never get linked. We've had a handful of social logins (and thus Django accounts created) by people that are not even in our country.

djnro/templates/front/management.html
```diff
@@ -17,12 +17,14 @@
     <h5>AAI Federation</h5>
     <hr>
     <p>{% blocktrans %}Authentication and authorization are carried out through a{% endblocktrans %} <a href="http://shibboleth.net/products/service-provider.html">Shibboleth SP</a>.</p>
-    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} <a href="http://aai.grnet.gr/documentation">{% trans "policy and procedures documentation" %}</a> {% trans "provided by the" %} {{ FEDERATION_NAME }}:
+    <p>{% blocktrans %}The following attributes are required for administrators and must be released by their home IdPs to the SP{% endblocktrans %} {% trans "according to the" %} {% if FEDERATION_DOC_URL %}<a href="{{ FEDERATION_DOC_URL }} ">{% endif %}{% trans "policy and procedures documentation" %}{% if FEDERATION_DOC_URL %}</a>{% endif %} {% trans "provided by the" %} {{ FEDERATION_NAME }}:
```
zmousm, line 20: I tend to prefer using whole sentences in blocktrans blocks, moving out of the way any embedded markup/variables, as I believe that might lead to more concise translation.
But no need to change something for now.

zmousm:
> Registration after invitation (like CAT) would cut down on the number of spurious social registrations that never get linked. We've had a handful of social logins (and thus Django accounts created) by people that are not even in our country.

We have seen similar behavior. I have reasons to believe that some users actually try to register because they think that's how they'll get to use eduroam... I have an idea about something that might help with such an "illusion", but let me not drift further off-topic :)

Thanks, merging.

PR #56, opened 2018-07-23
Django 1.11 - another upgrade to LTS Django

vladimir-mencl-eresearch:
Hi @zmousm ,

Time flies fast - Django 1.8 went EOL this April, so I've done upgrade to the next LTS - which is Django 1.11.

This time, it wasn't as disruptive as before.

Can you please have a look and merge?

Thanks a lot in advance!

Cheers,
Vlad


vladimir-mencl-eresearch:
I've just rebased on master and resolved conflicts (in `edumanage/views.py` - conflicting with ab6d29a

vladimir-mencl-eresearch:
Hi @zmousm, documentation added - now over to you for review.

edumanage/context_processors.py
```diff
@@ -1,4 +1,6 @@
 from django.conf import settings
+from django.core.urlresolvers import reverse
+import edumanage
```
cangus: A fully qualified import is probably preferable here, `edumanage.views.set_language` is used below and is probably working due to to it being imported elsewhere. Better to be explicit.

djnro/settings.py
```diff
@@ -77,28 +77,33 @@
...
+                'django.template.context_processors.request',
+                'edumanage.context_processors.country_code',
+                'edumanage.context_processors.cat_instances',
+                'edumanage.context_processors.manage_login_methods',
+                'edumanage.context_processors.reverse_urls',
+                'social.apps.django_app.context_processors.backends',
+                'social.apps.django_app.context_processors.login_redirect',
+            ],
+        },
+    },
```
cangus, line 105: Trailing comma's could be removed - they are only necessary for single element tuples to be interpreted correctly.

vladimir-mencl-eresearch: Agreed with @cangus to keep the trailing comma to avoid spurious (git) changes on the last line when adding another line - and to stay consistent with the rest of the code.

edumanage/views.py
```diff
@@ -2654,3 +2654,10 @@ def settings_dict_get(setting, *keys, **opts):
     if dct == {}:
         dct = opts.get('default', None)
     return dct
+
+# utility function to merge two dictionaries
+def merge_dicts(dict1, dict2):
```
cangus, line 2659: This could be moved to a more general utility module.
Also, could potentially be simplified to `return dict(itertools.chain(dict1.items(), dict2.items()))` (requires importing itertools)

edumanage/context_processors.py
vladimir-mencl-eresearch: Thanks, just pushed in a commit changing this to fully qualified import.

docs/installation/upgrading-from-1.1.md
```diff
@@ -0,0 +1,49 @@
+# Upgrading DjNRO from 1.1 to 1.2 or later
```
cangus, line 1: It might be worth mentioning the version that the migration is targeting.
This may reduce confusion later after several versions have passed and the context of the current master is lost

cangus:
Looks clean to me, a couple of tiny things mentioned above.

docs/installation/upgrading-from-1.1.md
vladimir-mencl-eresearch: This was written with the intention that after this PR gets merged, DjNRO 1.2 would get released.   So this doc assumes the previous migration [upgrading-from-1.0.md](upgrading-from-1.0.md) would be already applied and this would be next.

The document header says "from 1.1 to 1.2" (though filename only says "from 1.1").

How would you like to see the doc changed?

edumanage/views.py
vladimir-mencl-eresearch: Agreed with @cangus to keep this as-is - there are already other utility functions at the end of `edumanage/views.py` and as `merge_dicts` is only used from within this module, it makes the code more readable this way.

docs/installation/upgrading-from-1.1.md
cangus: I misread this as upgrading from any version between 1.1 and 1.2 to the current version.
I see what is intended now. No issue.

cangus:
All comments have been addressed - looks good to merge

djnro/settings.py
zmousm: Trailing commas are a good thing in this case, per PEP8.

zmousm (commented):
*moved inline*

accounts/migrations/0001_initial.py
```diff
@@ -20,7 +20,7 @@ class Migration(migrations.Migration):
...
                 ('last_login', models.DateTimeField(default=django.utils.timezone.now, verbose_name='last login')),
                 ('is_superuser', models.BooleanField(default=False, help_text='Designates that this user has all permissions without explicitly assigning them.', verbose_name='superuser status')),
-                ('username', models.CharField(error_messages={'unique': 'A user with that username already exists.'}, max_length=255, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.', 'invalid')], help_text='Required. 255 characters or fewer. Letters, digits and @/./+/-/_ only.', unique=True, verbose_name='username')),
+                ('username', models.CharField(error_messages={'unique': 'A user with that username already exists.'}, max_length=255, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.')], help_text='Required. 255 characters or fewer. Letters, digits and @/./+/-/_ only.', unique=True, verbose_name='username')),
```
zmousm: Could you give an example for the spurious migration prevented by the change to `django.core.validators.RegexValidator` `code` argument?

edumanage/views.py
zmousm: It's not a big deal but I would probably opt for changing `base_response()` to accept a dict as a second argument, to be merged with the context it produces.

Also +1 for `itertools.chain()`.

accounts/migrations/0001_initial.py
```diff
@@ -20,7 +20,7 @@ class Migration(migrations.Migration):
                 ('password', models.CharField(max_length=128, verbose_name='password')),
                 ('last_login', models.DateTimeField(default=django.utils.timezone.now, verbose_name='last login')),
                 ('is_superuser', models.BooleanField(default=False, help_text='Designates that this user has all permissions without explicitly assigning them.', verbose_name='superuser status')),
-                ('username', models.CharField(error_messages={'unique': 'A user with that username already exists.'}, max_length=255, validators=[django.core.validators.RegexValidator('^[\\w.@+-]+$', 'Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.')], help_text='Required. 255 characters or fewer. Letters, digits and @/./+/-/_ only.', unique=True, verbose_name='username')),
+                ('username', models.CharField(error_messages={'unique': 'A user with that username already exists.'}, max_length=255, validators=[django.contrib.auth.validators.ASCIIUsernameValidator()], help_text='Required. 255 characters or fewer. Letters, digits and @/./+/-/_ only.', unique=True, verbose_name='username')),
```
zmousm, line 23: Ditto. An example would help understand how a spurious migration is mitigated.

extras/servdata_consumer.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, src):
             if resp.status_code > 304 or not resp.ok:
                 exit_with_error("Fetch failed from %s" % src)
             self.rawdata = resp.text
+        elif self.src == "-":
```
zmousm, line 45: This was probably out of scope and could have been merged separately.

accounts/migrations/0001_initial.py
vladimir-mencl-eresearch: Hi @zmousm 

When I revert this commit and run `./manage.py makemigrations`, I get:
```
Migrations for 'accounts':
  accounts/migrations/0005_auto_20190225_1059.py
    - Alter field username on user
```
with `accounts/migrations/0005_auto_20190225_1059.py` having:
```
# -*- coding: utf-8 -*-
# Generated by Django 1.11.20 on 2019-02-24 21:59
from __future__ import unicode_literals

import django.contrib.auth.validators
from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('accounts', '0004_alter_user_last_login_null'),
    ]

    operations = [
        migrations.AlterField(
            model_name='user',
            name='username',
            field=models.CharField(error_messages={'unique': 'A user with that username already exists.'}, help_text='Required. 255 characters or fewer. Letters, digits and @/./+/-/_ only.', max_length=255, unique=True, validators=[django.contrib.auth.validators.ASCIIUsernameValidator()], verbose_name='username'),
        ),
    ]
```
which does not really change anything (no changes to the DB), but is recorded as a new migration, re-iterating all of the user model properties.  And it got triggered by the properties of UserModel in our copy not exactly matching what upstream evolved to.  So that's why I changed our copy of the user model to match upstream.

The other commit prevents pretty much exactly the same thing - both had to be in to make our model exactly match upstream.

extras/servdata_consumer.py
vladimir-mencl-eresearch: Yes, agree with that - but I did not want to end up with too many (long-outstanding) branches and decided there would be no harm in pushing this change in here....

accounts/migrations/0001_initial.py
vladimir-mencl-eresearch: see my other response below

docs/installation/upgrading-from-1.1.md
```diff
@@ -0,0 +1,49 @@
...
+
+The `social-auth-core` package has kept the same sequence of database migration steps, but because of the rename, Django would get confused and would try applying the migrations from start ... which would fail, because the tables already exist.
+
+Apply the following workaround: change the package name in the migrations history to make Django Migrations think these have been applied under the new package name.   (Because of the hierarchy used in the old package versions, the migrations were recorded as belonging to package `default`).  Run `./manage.py shell` and paste in the following code:
+
+    from django.db.migrations.recorder import MigrationRecorder
+    for m in MigrationRecorder.Migration.objects.filter(app='default'):
+        if not MigrationRecorder.Migration.objects.filter(app='social_auth', name=m.name):
+            print("Recording migration default:%s as social_auth:%s" % (m.name, m.name))
+            m_social=MigrationRecorder.Migration(app='social_auth', name=m.name)
```
zmousm, line 41: I'm not sure I understand why we would need to manually update applied migrations with `MigrationRecorder`, considering that PSA added the `replaces` attribute in https://github.com/omab/python-social-auth/commit/361f42235fe51f422af029f325d041407a16036c for migrations affected by the django app rename (`default` to `social_auth`, and this was carried over to `social_django`). Did you find this did not work for some reason?

accounts/migrations/0001_initial.py
zmousm: I see. What is even worse is that the username validator also changes based on whether it is running on Python 3...

While such migrations are noop and this is mostly an annoyance, I'm not sure what the most appropriate approach would be here, especially considering the runtime aspect above.

I need to investigate further...

zmousm: So yes the need to adapt to Django's changes to `django.contrib.auth.models.AbstractUser` is probably necessary; it would seem to be the price we pay for extending a complete class rather than `AbstractBaseUser`.

However I believe we should tweak the migration like https://github.com/django/django/commit/526575c64150e10dd8666d1ed3f86eedd00df2ed so as to avoid more spurious migrations sparked by a change in the run-time environment (PY3, eventually). I will submit a PR for that.

edumanage/context_processors.py
```diff
@@ -26,3 +28,6 @@ def cat_instances(context):
 
 def manage_login_methods(context):
     return {'MANAGE_LOGIN_METHODS': settings.MANAGE_LOGIN_METHODS}
+
+def reverse_urls(context):
```
zmousm, line 32: Even if the dotted path syntax was only used in one case, I think that using context manager(s) for this job is not a good idea, if nothing else because it is not generic -- you'd need one for each URL! I think we could just use a named URL. I will send a PR for that.

docs/installation/upgrading-from-1.1.md
vladimir-mencl-eresearch: Hi @zmousm , I also found the "replaces" annotation ... but wasn't working for me, and Django was trying to apply the new chain of migrations from scratch ... and was failing on that (that part is to be expected).

I'm not exactly sure why the "replaces"  annotation wasn't applying ... maybe because the original application did not load anymore?

I could do some experiments to confirm this - but I added it into the documentation because otherwise the migrations were failing...

Cheers,
Vlad

zmousm: I had a closer look at the issue, spending many hours navigating the infinite complexity of Django in order to understand the cause.

In my experience the problem is such a backtrace while running migrations:

```
  Applying social_django.0004_auto_20160423_0400...Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 204, in handle
    fake_initial=fake_initial,
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 115, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 145, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/db/migrations/executor.py", line 244, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/db/migrations/migration.py", line 119, in apply
    operation.state_forwards(self.app_label, project_state)
  File "/srv/venv/djnro_django111/local/lib/python2.7/site-packages/django/db/migrations/operations/fields.py", line 201, in state_forwards
    state.models[app_label, self.model_name_lower].fields
KeyError: ('social_django', u'usersocialauth')
```

Searching about that will give you https://github.com/python-social-auth/social-app-django/issues/28, where the suggested workaround is to first upgrade to PSA 0.2.21, as an intermediate step. In our case you did this in 9ff2496, where the Django 1.11 upgrade is already under way. I think it should have been committed (and tagged) before this PR. We could then guide users to first checkout that tag, `pip install -r ...` and run migrations, per the suggested workaround, before going further. Nevertheless we can still instruct them how to do this manually.

Your suggested workaround involving `MigrationRecorder` also works, of course, and I admit it is more simple. In principle however I think it is best to avoid juggling with Django internals for user-oriented guidelines.

I have updated the upgrading guide, providing an explanation for the issue as well as both workarounds, and pushed this to #58.

zmousm: The actual reason for this backtrace is as follows:

When jumping over PSA 0.2.21, the migrations for `social_auth` are not recorded in the database. `social_django` migrations are declared as _replacements_ for respective migrations in both `default` and `social_auth` apps. Django however requires that _**all** replacement targets are marked as applied for the replacing migration to be marked as applied_. See https://github.com/django/django/blob/331d76528154407927f88759c9e520494e29b914/django/db/migrations/loader.py#L229. Failing this, `social_django` models are excluded from the _project state_, and this leads to the `KeyError` in question. The connection between the two seems quite loose, admittedly, but I think one could even argue this constitutes a bug (or two) in Django.

vladimir-mencl-eresearch: Hi @zmousm , thanks a lot for this code deep dive.
Yes, you are right - I initially started with updating to last PSA (0.2.21).
In the end, I decided for the workaround I put into the documentation as easier to follow (though I'm neither a fan of touching Django internals either).  Happy with your other workaround as well.
Thanks again for all the hard work put into this!
Cheers,
Vlad

PR #57, opened 2018-10-17
CATv2 API changes

ghalse:
With the release of CAT 2.0-beta1 comes a significant change in the CAT [admin API](https://github.com/GEANT/CAT/blob/master/Changes.md). This change is not [backwards compatible](https://wiki.geant.org/display/H2eduroam/A+guide+to+eduroam+CAT+2.0+and+eduroam+Managed+IdP+for+National+Roaming+Operator+administrators#AguidetoeduroamCAT2.0andeduroamManagedIdPforNationalRoamingOperatoradministrators-UI-lessAutomatedManagement:theAdminAPI(2.0)), and the old API version is no longer supported in production.

DjNRO abstracts its CAT API calls into a `CatQuery` class, but the parameters passed to functions in this class are explicitly derived from v1 of the CAT API. This change introduces support for the CATv2 admin API while preserving the DjNRO `CatQuery` abstraction interface. By doing this, we avoid making changes to other parts of the DjNRO code base at the expense of slightly complicating the `CatQuery` class.

Part of the reason I've elected to do things in this way is that the `CatQuery` class contains functions and static methods that are not called from elsewhere within the published DjNRO code base, making it unclear as to why they exist -- and leading to the assumption that they're used by other, unpublished utilities.

### DIFF
--- a/utils/cat_helper.py
+++ b/utils/cat_helper.py
@@ -91,8 +91,7 @@ class CatQuery(object):
                 params[id]['NAME'] = kwargs[k]
             elif k.startswith('value[S'):
                 if k.endswith('-lang]'):
-                    if kwargs[k] != 'C':
-                        params[id]['LANG'] = kwargs[k]
+                    params[id]['LANG'] = kwargs[k]
                 else:
                     params[id]['VALUE'] = kwargs[k]
         return filter(None, params)
--- a/utils/cat_helper.py
+++ b/utils/cat_helper.py
@@ -41,63 +41,114 @@ class CatQuery(object):
         except (XMLSyntaxError, AssertionError):
             return objectify.Element('Dummy')
 
+    """ Send a CATv2 API request """
+    def v2api_request(self, verb, params):
+        apiRequest = {
+            "ACTION": verb,
+            "APIKEY": self.key,
+            "PARAMETERS": params
+        }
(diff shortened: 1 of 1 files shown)


zmousm:
Sorry for the very very long delay.

To quote the author: the CAT version 1 admin API was "a terrible ad-hoc hack". You would send it form data and it would respond with XML. The input was somewhat peculiar (quoting public documentation, again), as one was expected to embed indexing and data type info in the name of the form-field. The output was some times invalid XML (e.g. for some errors). DjNRO tried to cope with that, but it had limited scope so it made no sense to fully abstract all aspects of such a "draft" API in `CatQuery`.

> Part of the reason I've elected to do things in this way is that the CatQuery class contains functions and static methods that are not called from elsewhere within the published DjNRO code base, making it unclear as to why they exist -- and leading to the assumption that they're used by other, unpublished utilities.

I didn't understand this statement; what were the functions and static methods you referred to?

Anyway, thank you very much for picking this up. I pushed a couple of tweaks/fixes and rebased as necessary after the Django 1.11 update. I would have preferred to also maintain compatibility with the old CAT API. In order to achieve that, I would reverse the deuglify processing you introduced, so that input to `CatQuery` methods aligns with the new API, but it may also be translated to parameters for the old (and more limited) API. I want to revisit this, but it is not a topmost priority.

PR #61, opened 2019-05-10
connect: tweak template to enable overriding selectors in the institution list

vladimir-mencl-eresearch:
Hi @zmousm ,

Thanks for the detailed suggestions in #60 - works for me perfectly.
I've now added the suggested `cat_institution_selector` template block (and also bumped font-awesome to 4.7.0).

I was looking for a more suitable icon to use in our setup - I found https://fontawesome.com/icons/user-cog?style=solid, but that would require font-awesome 5, and that might be too much of an overhaul.

I'm in the end using "cogs" icon in our setup - that provides a hint to the users that this might be setting their device up.

Happy to contribute the icon bit if you in the end find it useful - but also happy just having it in a local template.

Leaving this one for you to look at (but understand this would be also OK to merge right away...)

Cheers,
Vlad

### DIFF
--- a/djnro/templates/base.html
+++ b/djnro/templates/base.html
@@ -13,7 +13,7 @@
 <link rel="apple-touch-icon-precomposed" sizes="114x114" href="{% static 'img/edu114-icon.png' %}">
 <link rel="apple-touch-icon-precomposed" sizes="72x72" href="{% static 'img/edu72-icon.png' %}">
 <link rel="apple-touch-icon-precomposed" href="{% static 'img/edu-icon.png' %}">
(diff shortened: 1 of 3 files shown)


zmousm:
I'm happy to merge once the inline comment gets resolved. I do wonder what regressions the FA bump may bring but it's hard to tell anyway (yes this would be more difficult for a FA 5 bump).

zmousm:
> I'm happy to merge once the inline comment gets resolved

I did this myself. I hope you're OK with the wording. I would prefer to amend 996ee45 rather than push a new commit.

vladimir-mencl-eresearch:
Thanks @zmousm  -  all happy with you changing the wording on my behalf.

PR #63, opened 2019-05-29
selectinst: return 400 Bad Request for method!=GET

vladimir-mencl-eresearch:
... otherwise, the code returns None, which causes an exception in the @never_cache decorator.

Replace an embarassing stack trace / 500 error page with a more graceful error message.


vladimir-mencl-eresearch:
Trivial fix, merging

PR #67, opened 2019-08-21
admin: fix TinyMCE reverse URL

vladimir-mencl-eresearch:
tinymce/urls.py defines a name to use for the flatpages_link_list view - so use
this name in the reverse() call.


vladimir-mencl-eresearch:
Trivial fix, successfully tested both on Python2 and Python3, merging.

PR #70, opened 2019-12-16
eduroam database v2 adaptation

### LINKED ISSUE #62: eduroam database version 2 adaptation
## eduroam db v2 schema changes

[realm]
* Changed: realm.xml renamed to ro.xml (and ro.json)
* Added: ROid: string, provided by eduroam OT
* Added: stage: 0=preproduction/test, 1=active (default)
* Changed: address_street, address_city -> multilingual
* Added: coordinates: string: lon,lat[,alt] (optional)
* Added: contact_type: 0=person (default), 1=service/department
* Added: contact_privacy: 0=private (default), 1=public
* Removed: stype

[institution]
* Added: instid: string, provided by NRO
* Added: ROid: string, provided by eduroam OT
* Changed: type: enumeration: int -> string
* Added: stage: 0=preproduction/test, 1=active (default)
* Renamed: org_name -> inst_name
* Changed: address_street, address_city -> multilingual
* Added: coordinates: string: lon,lat[,alt] (optional)
* Added: inst_type: string: IEEE 802.11-2012, clause 8.4.1.34 Venue Info (note: this is a pair of numbers separated with comma (,))
* Added: contact_type: 0=person (default), 1=service/department
* Added: contact_privacy: 0=private (default), 1=public
* Removed: realmid

[service_loc]
* Added: instid: string, provided by NRO
* Added: ROid: string, provided by eduroam OT
* Added: locationid: string, provided by NRO
* Added: coordinates: string: lon,lat[,alt][;lon,lat[,alt];...]
* Added: stage: 0=preproduction/test, 1=active (default)
* Added: type: 0=single spot (default), 1=area, 2=mobile
* Changed: address_street, address_city -> multilingual
* Added: location_type: string: IEEE 802.11-2012, clause 8.4.1.34 Venue Info (note: this is a pair of numbers separated with comma (,))
* Changed: enc_level: optional
* Removed: port_restrict
* Removed: transp_proxy
* Removed: IPv6
* Removed: NAT
* Added: tag: string, comma-separated, allowed values: port_restrict, transp_proxy, IPv6, NAT, HS2.0
* Changed: AP_no: blank if unknown
* Added: wired_no: number of sockets, blank if unknown
* Added: availability: 0=no restrictions (default), 1=physical access restrictions
* Added: operation_hours: string: free text, opening hours in local time
* Removed: institutionid
* Removed: longitude
* Removed: latitude
* Removed: wired

## eduroam db v2 schema caveats

* Intitution type: schema switched from integer to string; we can not extend the tuple (ERTYPES)
  we use for mapping because django does not allow tuples with len > 2 to be used for choices
* Address: we want to use lang on the whole object, yet both json/xml schema specify lang on each
  of the street, city fields, which does not make sense; we shall have to transform as necessary
* Should service_loc type be derived from number of coordinates? 0 mobile, 1 spot, >1 area
  However it would never yield 0 (mobile) because service_loc requires number of coordinates >= 1
  In other words: does it make sense for a service_loc type=mobile to have any coordinates?
* No explicit versioning scheme defined for eduroam database requests/endpoints (only implicitly for
  realm->ro and json endpoints); if multiple versions are supported (backwards compat), a setting must define which version should be rendered

## coordinates database modeling

* coordinates uniformly represented on all 3 tables: serialized tuples of (lon, lat[, alt]) joined with ;
  * realm: 0 or 1 tuples
  * institution: 0 or 1 tuples
  * service_loc: >= 1 tuples
* how to model? options:
  * custom CharField (like MultiSelectField) (de-)serializing from/to a list of objects
    * simple but inefficient (storage, queries)
  * GeoDjango: PointField and LineStringField (corresponding to polygon in the eduroam db schema def)
    * most efficient but GeoDjango out of scope currently (overkill)
  * separate model (table) with relation:
    * one of:
      * ForeignKeyField and *sorted* ManyToManyField
        * m2m sorting can be:
          * implemented manually through a field in the intermediate table
          * provided by django-sortedm2m along with form widgets, at the expense of adding a dependency
        * sorting also needed for other m2m relations
      * GenericRelation (django.contrib.contenttypes.fields)
        * polygon point sorting must be implemented manually
        * no intermediate table per m2m relation
        * no other particular reason (e.g. polymorphism) to prefer (more complicated) contenttypes framework?
    * db-persisted as separate fields -> somewhat efficient for querying; less efficient for storage (due to sorting records for all tuples would be updated upon save)
    * (de-)serialization and validation through model instance (rather than field) method(s)
* polygon validation (non self-intersecting) non trivial without GeoDjango or some GIS library (Shapely)
* compat considerations:
  * service_loc shall be restricted initially to exactly one set of coordinates
  * service_loc latitude, longitude properties should be preserved as proxies to the actual fields

## djnro models changes

* Add enums/choices tuples
  * PRODUCTION_STATES
  * CONTACT_TYPES
  * CONTACT_PRIVACY
  * LOCATION_TYPES
  * PHYSICAL_AVAILABILITY_STATES
  * LOCATION_TAGS
  * ERTYPES_DB ?

### Contact
* Add type PositiveIntegerField (choices=CONTACT_TYPES, default=0)
* Add privacy PositiveIntegerField (choices=CONTACT_PRIVACY, default=0)

### Address_i18n (new model)
* street
* city
* lang ...

### Realm
* Add roid CharField
* Add stage PositiveIntegerField (choices=PRODUCTION_STATES, default=1)
* coordinates? (see discussion)
* Add address GenericRelation(Address_i18n) field
* Remove stype field

### Institution
* Add instid UUIDField
* Add stage PositiveIntegerField (choices=PRODUCTION_STATES, default=1)
* Rename org_name -> inst_name

### InstitutionDetails
* Add address GenericRelation(Address_i18n) field
* coordinates? (see discussion)
* Add venue_info Charfield (validators=[validate_comma_separated_integer_list,validate_venue_info])
* Remove number_user field
* Remove number_id field

### ServiceLoc
* Add slocid UUIDField
* coordinates? (see discussion)
* Add stage PositiveIntegerField (choices=PRODUCTION_STATES, default=1)
* geo_type field? derived from number of coordinates? default=mobile?
* Add address GenericRelation(Address_i18n) field
* Add venue_info CharField (validators=[validate_comma_separated_integer_list,validate_venue_info])
* enc_level already accepts blank/null?
* tags: add MultiSelectField like enc_level
* AP_no: change blank=True, null=True ?
* Add wired_no PositiveIntegerField? blank=True, null=True?
* Add physical_avail PositiveIntegerField (choices=PHYSICAL_AVAILABILITY_STATES, default=0)
* Add operation_hours CharField

### RealmData
* remove model

## djnro data migrations
* address_street, address_city -> create address lang=en for each of
  * Realm
  * InstitutionDetails
  * ServiceLoc
* ServiceLoc: port_restrict, transp_proxy, IPv6, NAT if yes -> tags
* ServiceLoc: latitude, longitude -> coordinates

### Realm
* Remove address_street field
* Remove address_city field

### InstitutionDetails
* Remove address_street field
* Remove address_city field

### ServiceLoc
* Remove address_street field
* Remove address_city field
* Remove port_restrict field
* Remove transp_proxy field
* Remove IPv6 field
* Remove NAT field
* Remove latitude, longitude? Replace with property methods?

TBA: forms, views, templates

zmousm:
Implement #62 

~~WIP. Models and migrations are mostly done. Still pending forms, views, templates management commands.~~ Now complete.

Important notes, caveats:
* `UUID` is enforced for `Institution.instid`, `ServiceLoc.locationid`. Such fields are not editable through forms, while `parse_institution_xml` defaults to deriving a `UUID` by means of MD5 hashing, if the input is unsuitable.
* `Coordinates` are modeled as m2m for `ServiceLoc`, but only a single set is supported currently. This is enforced by a signal receiver. Longitude, latitude `ServiceLoc` cached properties work as a proxy, together with a related mechanism in `ServiceLocForm`. On the other hand, `parse_institution_xml` retains the first set of coordinates it encounters in a location.
  * Therefore altitude is currently ignored.
* `Coordinates` are currently ignored in institution edit form, nor are they otherwise presented anywhere, while they are processed by `parse_institution_xml` as expected.
* Data migration from `ServiceLoc.wired` to `ServiceLoc.wired_no` is driven by `settings.SERVICELOC_DERIVE_WIRED_NO`, which defaults to mapping `True` to 42.
* Compatibility with all eduroam database versions is retained throughout db views and `parse_institution_xml`. For the latter the version may be provided through a command-line option. For the former the requested version is inferred by considering path, query parameter or resource (realm vs. ro). If no version is specified/hinted, a default is used, which, as of now, is version 2.
  * This should probably be documented in release/upgrade notes.

@vladimir-mencl-eresearch,@c00kiemon5ter, @ioparaskev please start reviewing


vladimir-mencl-eresearch:
Hi @zmousm , thanks a lot for this work!

I'm just about to go on Christmas leave (for southern hemisphere/New Zealand, these are the main summer holidays) and I'll be back on deck from January 13th.

I'll see if I manage to review anything in the last ~ hour I have in the office, but I'm afraid in won't be much - I promise to look at this once I'm back in the week of January 13th.

Cheers,
Vlad

zmousm:
No worries @vladimir-mencl-eresearch, happy holidays!

edumanage/models.py
```diff
@@ -21,109 +42,157 @@ class MultiSelectFormField(forms.MultipleChoiceField):
...
-        #             % (apnumber(self.max_choices), pluralize(self.max_choices)))
-        return value
-
+        supercls = super(MultiSelectFormField, self)
+        # remove TypedChoiceField extra args
+        supercls_init_args = getargspec(supercls.__init__).args[1:]
+        for key in list(kwargs):
+            if key not in supercls_init_args:
+                del kwargs[key]
+        supercls.__init__(*args, **kwargs)
```
ioparaskev: maybe instead of iterating and deleting keys from the dict, you could achieve the same thing with:
```python
supercls.__init__(*args, **{key,val for (key,val) in kwargs.items() if key in supercls_init_args})
```

edumanage/models.py
```diff
@@ -21,109 +42,157 @@ class MultiSelectFormField(forms.MultipleChoiceField):
...
+        self.separator = kwargs.pop(
+            'separator', self._separator_default)
+        super(MultiSelectField, self).__init__(*args, **kwargs)
 
-    def get_choices_default(self):
-        return self.get_choices(include_blank=False)
+    def deconstruct(self):
+        name, path, args, kwargs = super(
+            MultiSelectField, self).deconstruct()
+        if self.separator != self._separator_default:
```
ioparaskev, line 70: since we always want to have the `kwargs['seperator'] == self.seperator` why not make the assignment in the first place? I'm not sure there's a point having the `if` statement here unless I'm missing something

edumanage/models.py
zmousm: Sure. I thought it was better readable like that.

edumanage/models.py
zmousm: Django docs [suggest the deconstructor should only return kwargs that don't carry the default value](https://docs.djangoproject.com/en/1.11/howto/custom-model-fields/#field-deconstruction) (as they are deemed redundant otherwise).

edumanage/models.py
ioparaskev: main reason I proposed this was to avoid having `del`

vladimir-mencl-eresearch:
Hi @zmousm ,

Great job.  I've just gone through all the commits.  I'll comment inline, but it's mostly requests for clarification / more detailed documentation - no real issues discovered, looks good to me.

Cheers,
Vlad

edumanage/models.py
```diff
@@ -100,30 +101,71 @@ def value_to_string(self, obj):
         value = self._get_val_from_obj(obj)
         return self.get_db_prep_value(value)
 
-ERTYPES = (
-    (1, 'IdP only'),
-    (2, 'SP only'),
-    (3, 'IdP and SP'),
+# https://www.djangosnippets.org/snippets/2402/
+def get_namedtuple_choices(*choices_tuples):
```
vladimir-mencl-eresearch, line 185: Hi @zmousm ,

I think it would be useful to document the required structure of the choices_tuples passed to this call.

From the code here, I can each tuple in the choices should start with an identifier, but I see the structure then quite varies:
* ER_TYPES has a numeric ID, others don't.
* ER_TYPES has has a short handle *after* a descriptive name, others reverse this order.

I think it would be good to define the structure to be passed - and then adhere to it.

edumanage/models.py
```diff
@@ -100,30 +101,71 @@ def value_to_string(self, obj):
...
+            return attr[0]
+    return Choices._make(vals)
+
+_ERTYPES = (
+    ('IDP', 1, 'IdP only', 'IdP'),
+    ('SP', 2, 'SP only', 'SP'),
+    ('IDPSP', 3, 'IdP and SP', 'IdP+SP'),
+)
+ERTYPES = get_namedtuple_choices(*_ERTYPES)
+def get_ertype_string(ertype):
```
vladimir-mencl-eresearch: This function should document the input parameter.  I assume it is the numeric ID, but it should be documented.

edumanage/models.py
```diff
@@ -222,6 +320,27 @@ def __str__(self):
         return self.url
 
 
+@python_2_unicode_compatible
+class Address_i18n(models.Model):
```
vladimir-mencl-eresearch, line 406: I understand the structure is driven by what eduroam.org asked for ... but shouldn't an address also specify a Country?
Street and City does not look sufficient to me in the global context....

edumanage/models.py
```diff
@@ -520,6 +716,12 @@ class InstitutionDetails(models.Model):
         blank=True,
         help_text=_('The primary, registered domain name for your institution, eg. example.com.<br>This is used to derive the Operator-Name attribute according to RFC5580, par.4.1, using the REALM namespace.')
     )
+    venue_info = models.CharField(
+        max_length=7,
+        blank=True,
+        validators=[validate_venue_info],
+        db_column='inst_type'
+    )
```
vladimir-mencl-eresearch, line 842: Why is the field called`venue_info` but stored in db_column `inst_type` ?  That just looks like unnecessary complexity - can't we just use one name for both?

I expect there is a reasonable answer to this - just asking because it struck me as odd.

edumanage/models.py
```diff
@@ -427,30 +546,64 @@ class ServiceLoc(models.Model):
...
+        default=GEO_TYPES.SPOT,
+        db_column='type'
+    )
     # TODO: multiple names can be specified [...] name in English is required
     loc_name = fields.GenericRelation(Name_i18n)
-    address_street = models.CharField(max_length=96)
-    address_city = models.CharField(max_length=64)
+    # TODO: multiple addresses can be specified [...] address in English is required
+    address = fields.GenericRelation(Address_i18n)
+    venue_info = models.CharField(
```
vladimir-mencl-eresearch, line 697: Same comment as for InstitutionDetails.venue_info:

> Why is the field called`venue_info` but stored in db_column `inst_type` ?  That just looks like unnecessary complexity - can't we just use one name for both?
> 
> I expect there is a reasonable answer to this - just asking because it struck me as odd.

edumanage/models.py
```diff
@@ -184,6 +264,24 @@ class Meta:
         verbose_name_plural = "Contacts"
 
 
+@python_2_unicode_compatible
+class Coordinates(models.Model):
```
vladimir-mencl-eresearch, line 342: I think it would be helpful to document the units usedhere.

I assume it's degrees as float for lat/long - and meters for alt ... but I think it should better be documented.

vladimir-mencl-eresearch (changes requested):
And that's it.

Note that I did not review: the changes to `MultiSelectField` / `MultiSelectFormField` (a bit too much outsideof  my area) - but it appears to have been reviewed extensively by @ioparaskev.

So once more, congratulations on the job well done.

Cheers,
Vlad

edumanage/models.py
zmousm: Hello @vladimir-mencl-eresearch,

the structure is the sequence of iterables that [Django accepts for choices](https://docs.djangoproject.com/en/1.11/ref/models/fields/#choices), and in each of them the two items are preceded by the [field name for namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple). _ERTYPES is special in that each iterable also includes at the end the string value, which is new in eduroam db v2, so as to help with the mapping. However any such further items are ignored by `get_namedtuple_choices`.

Coming up, I will be adding some documentation.

edumanage/models.py
zmousm: Alright

edumanage/models.py
zmousm: One might argue this can be derived considering the RO to which the particular object is linked and the country (area) they serve. I do get your point though, yet I am focused on covering eduroam db v2 schema changes in this (quite large) change set, and this can be added later.

edumanage/models.py
zmousm: The eduroam db spec, both the abstract version and the XML (JSON) schema, defines some quite generic field names, which felt too confusing (or even technically inappropriate) to use for our Django model fields. We have always followed the convention of using the actual name in the spec for `db_column`, if at all possible, as this is hidden from casual usage, but we still get to record the mapping in there. And now we will actually get to use this mapping, as you will see in upcoming commits. I do understand however that as new eduroam db versions arise, such a mapping mechanism is inadequate and should be replaced by something that also considers the EDB version. I could not tackle this at this stage, though.

edumanage/models.py
zmousm: See previous

edumanage/models.py
zmousm: Right. I will try to add some `help_text` in some fields.

edumanage/models.py
vladimir-mencl-eresearch: Sure, no problem.  I understand this is driven by the V2 structure and we can't change it easily from here - still wanted to raise it as an usability issue.  But that's not to be solved here, so I'm marking this part as resolved.

edumanage/models.py
vladimir-mencl-eresearch: Sure - also happy with this - glad to hear you did consider this carefully.

edumanage/models.py
vladimir-mencl-eresearch: Agreed.

edumanage/models.py
vladimir-mencl-eresearch: Agree that when I treat `ER_TYPES` as a separate special case, the structures look more uniform - a tuple of 3-tuples, each starting with an identifier.

Still, sometimes, the second field in each tuple (the internal value to store?) is a string, sometimes a number.  I assume the last field in each tuple is a description.

Having this documented would help a lot - thanks a lot in advance for that.

zmousm: > Still, sometimes, the second field in each tuple (the internal value to store?) is a string, sometimes a number. I assume the last field in each tuple is a description.

That is correct, per the Django spec.

zmousm:
I am now pushing the second round of commits (hopefully there won't be a third round).

The first 12 commits (prefixed with _fixup!_) are meant to be squashed into their preceding counterparts before merging. I am not doing this now so as to not confuse you and not force you to review everything all over again.

@vladimir-mencl-eresearch, @c00kiemon5ter, @ioparaskev please review.

I hope to find some time to update the high-level overview of the changes in the linked issue.

zmousm:
I pushed 4 more commits to fix errata

zmousm:
I should also note that I have not effectively tested `parse_institution_xml` as of yet...

edumanage/models.py
vladimir-mencl-eresearch: Thanks, looks all good now.

edumanage/models.py
vladimir-mencl-eresearch: Thanks for the patience with me - happy with just pointing to Django doc.

zmousm:
I am now pushing the final round of fixes, after testing `parse_institution_xml`.

zmousm:
Thank you for your efforts, I understand testing and/or reviewing is not easy for this particularly large change-set.

I will squash all _fixup!_ commits and merge this shortly due to (external) time constraints.

vladimir-mencl-eresearch:
Hi @zmousm ,

Thanks for all the effort put into it - and sorry about the slow action on my side.
I've been trying to go through the second batch of commits, but was not progressing much - and there was a lot to follow, sometimes out of my depth with Django quirks, sometimes just plain lots of changes.  But overall, it all looked reasonable.

If you'd like me to at least help testing, please let me know when ready to go (and which branch).

Cheers,
Vlad

zmousm:
Hi @vladimir-mencl-eresearch,

throughout this effort I tried to only touch code that had to be updated for EDB v2, yet avoid working around code that needed refactoring. DjNRO is an 8 year old project, so the latter was all over the place :) I am sorry if I made it harder for you.

What you can do is deploy the code in master (considering the notes/caveats provided previously) and check if it works out for any niche use cases you may have in mind -- maybe you will find a bug I didn't catch!

PR #71, opened 2020-01-13
extras/servdata_consumer.py: Properly handle pre-encoded Mako output

zmousm:
* `ServdataWriter` specifies `output_encoding` for Mako, thus `render()` returns PY2/PY3 `str`/`bytes`, i.e. already encoded output
* To properly handle this
  * Open output file in binary mode
  * Use `sys.stdout.buffer.write()` for PY3
* We can then remove the `six` dependency

### DIFF
--- a/extras/servdata_consumer.py
+++ b/extras/servdata_consumer.py
@@ -14,7 +14,6 @@ except ImportError:
 import requests
 from mako.template import Template
 from mako.lookup import TemplateLookup
-import six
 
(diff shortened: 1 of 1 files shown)


vladimir-mencl-eresearch (approved):
Hi @zmousm ,

Thanks - looks all good to me.  Nice cleanup that removes the workarounds I had put in.

Ready to merge from my point of view.

Cheers,
Vlad

PR #74, opened 2020-03-08
Fix EDB2 migrations

### LINKED ISSUE #73: Migration issue on PostgreSQL < 11
Hi @zmousm ,

I have successfully upgraded REANNZ DjNRO instance to 1.2, and I'm now trying to update the Docker-packaged version I created for the XeAP project - so that these deployments can also update to EDB2.

When testing the migration locally, I have `edumanage/migrations/0005_edb2_schema_part1.py` failing with: 
```
psycopg2.OperationalError: cannot ALTER TABLE "edumanage_serviceloc" because it has pending trigger events
```

This deployment is using PostgresQL 9.5.  I did not run into this with the REANNZ instance, which runs with PostgreSQL 11.  I have experimented with different version, PostgreSQL 10 has the issue as well - something changed between 10 and 11 that makes the problem go away.

I have found references to this issue in the Django Migrations documentation - https://django.readthedocs.io/en/1.11.x/ref/migration-operations.html#runpython explicitly says:

> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

From these pointers, I also found I can work around this problem by making the migration non-atomic:
```
 class Migration(migrations.Migration):

+    atomic = False
```
I don't know how many DjNRO deployments are out there, and whether it would be worth supporting older versions.  Also not sure how to weigh the risks associated with making the migration non-atomic.

For the XeAP deployment, I can just ship a patched version that would make this migration non-atomic.

I could also generalize this and make it conditional on "databases that support DDL transactions", where this issue might happen (because they combine schema alterations and RunPython code).

Something like
```
    atomic = django.db.connection.vendor not in ['postgresql', 'sqlite3']
```

What do you think?  Should I create a PR with this - or not worth it?

Cheers,
Vlad

zmousm:
* Make boolean, address, longitude/latitude fields (replaced in EDB2) nullable before removal
  * This is necessary so as to be able to perform a backwards migration, but it has an unfortunate implication: missing necessary data, it will not error as soon as the data migration happens but upon the nullable flag being reset
* Separate schema from UUID data backfilling in different migrations, so as to avoid a problem on _databases that do support DDL transactions_, such as PostgreSQL

Fix #73


vladimir-mencl-eresearch:
Thanks for the quick fix @zmousm  - tests all fine in my environment!

PR #75, opened 2020-03-09
Fix font-awesome tags

ghalse:
When font-awesome was upgraded in #61, some icons got left behind. While mostly cosmetic, this did create a functional problem with the "show password" link.

This replaces all of the legacy v3.2 `icon-` tags with their direct equivalents in font-awesome 4.7.0. While I was at it, I've also been through the code and found all the `fa-` tags and verified they still exist in 4.7.0.

Given that all font-awesome icons are now tagged with the `fa` class, this should make a future upgrade a bit simpler too.

### DIFF
--- a/djnro/static/js/showpass.js
+++ b/djnro/static/js/showpass.js
@@ -19,7 +19,7 @@ showPassword: function() {
 
 var input_password	= $(this);
 //create the icon and assign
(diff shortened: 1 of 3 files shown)


zmousm:
Thanks. I had noticed the discrepancy while housekeeping `services_edit.html` but I didn't get to fix things at the time.

PR #80, opened 2020-05-27
Limit participants display to active participants

ghalse:
The new data model includes a production state/stage of "preproduction/test". It probably doesn't make sense to include these in the public list of participants, and to limit the public view only to institutions that are marked as production-ready.

This change filters the /participants view accordingly.

### DIFF
--- a/edumanage/views.py
+++ b/edumanage/views.py
@@ -54,6 +54,7 @@ from edumanage.models import (
     Coordinates,
     ERTYPES,
     ERTYPE_ROLES,
+    PRODUCTION_STATES,
 )
 from .models import get_ertype_string
 from accounts.models import UserProfile
@@ -1669,7 +1670,8 @@ def api(request):
 
 @never_cache
 def participants(request):
-    institutions = Institution.objects.filter(institutiondetails__isnull=False).\
+    institutions = Institution.objects.filter(stage=PRODUCTION_STATES.ACTIVE,
+                                              institutiondetails__isnull=False).\
       select_related('institutiondetails')
     cat_instance = 'production'
     dets = []


vladimir-mencl-eresearch:
Hi @ghalse ,

Thanks for the contribution.  This looks very reasonable to me - and I've just tested and does exactly what it should.

@zmousm , OK to merge?

Cheers,
Vlad

zmousm:
I would prefer it if there was a selector to toggle filtering between active and preprod entitites.

In general this view (template) has been waiting for a large refactoring after _Connect (with CAT)_ was introduced, so as to zoom to a particular entity, better display more details, contacts etc. But it would eventually switch to client-side rendering, after the API endpoints are finally added.

If you do want to merge this in the meantime I would certainly not object, especially if a filter toggle could be added.

ghalse:
I'm not sure what you envision with a selector? As in a config flag in the config file? If so, I have no problem adding that.

My use case for this is to hide pre-production participants that are in live testing but still not publicly announced. In the particular scenario that resulted in this patch, it's a fairly high profile potential SP and we don't want to be responsible for leaking their participation publicly until they're ready. Thus any publicly accessible selector defeats this.

edumanage/views.py
```diff
@@ -1669,7 +1670,8 @@ def api(request):
 
 @never_cache
 def participants(request):
-    institutions = Institution.objects.filter(institutiondetails__isnull=False).\
+    institutions = Institution.objects.filter(stage=PRODUCTION_STATES.ACTIVE,
+                                              institutiondetails__isnull=False).\
```
zmousm, line 1674: Maybe you can fix the indentation/wrapping here.

zmousm: I will do it

zmousm:
Yes I meant a selector on the page to modify what is rendered, client-side (even css).

I get your point though, so I won't insist.

There are probably more places where such filtering may be likewise appropriate, so perhaps tweaking the manager may be handy, but perhaps we can revisit that later.

PR #81, opened 2020-06-12
Add missing EDB2 data

ghalse:
It seems that commit cda7f5758c8356d229229f60682904a478e9c897 missed adding the new type and privacy fields to contacts_edit.html


zmousm:
Thanks

PR #82, opened 2021-03-14
Fix: Closest point API: ignore individual point parsing errors

vladimir-mencl-eresearch:
Hi @zmousm ,

I hope you are doing well.

Last week, we ran into issues where the upstream KML file had unparsable data in latitude/longitude fields (coordinates accidentally expressed as `'lat': "57°0'56.", 'lng': "9°58'36."` - with unicode degree character!

DjNRO started emailing us stack-trace error messages for each hit to the closest point API - and users were getting 500 Internal Server error.

This PR is a work-around that catches `ValueError` within the loop, just skipping invalid entries.

I raised the issue with the eduroam OT - they were not able to find where that error came from, and it disappeared in the meantime.  I still have a copy of the invalid data ... but while it's not possible to track how it got into the KML file, it seems to be reasonable to have a safeguard against future invalid data.

Does this look OK to you to merge?

Cheers,
Vlad

### DIFF
--- a/edumanage/views.py
+++ b/edumanage/views.py
@@ -1867,8 +1867,14 @@ def closest(request):
             pointlat = i['lat']
             pointtext = i['text']
             plainname = i['name']
-            dLat = rad(float(pointlat) - float(lat))
-            dLong = rad(float(pointlng) - float(lng))
+
+            try:
+                dLat = rad(float(pointlat) - float(lat))
+                dLong = rad(float(pointlng) - float(lng))
+
+            except ValueError:
+                continue # shrug: parsing error
+
             a = math.sin(dLat / 2) * math.sin(dLat / 2) + math.cos(rad(lat)) *\
                 math.cos(rad(float(pointlat))) * math.sin(dLong / 2) *\
                 math.sin(dLong / 2)


edumanage/views.py
```diff
@@ -1862,28 +1862,31 @@ def closest(request):
...
-                    'plainname': plainname
-                }
+            try:
+                pointname = i['text']
+                pointlng = i['lng']
+                pointlat = i['lat']
+                pointtext = i['text']
+                plainname = i['name']
+                dLat = rad(float(pointlat) - float(lat))
+                dLong = rad(float(pointlng) - float(lng))
```
zmousm, line 1873: The exception catching block should be limited to float casting statements.

zmousm:
Hello @vladimir-mencl-eresearch

I saw an XML parsing error (perhaps related) around the same time:
```
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 94808, column 64
```

Ideally any such issues should be caught through input (XML schema) validation, but I agree such a safeguard is reasonable as well, in the short term.

Please see the inline comment.

edumanage/views.py
vladimir-mencl-eresearch: Hi Zenon, good catch - just adjusted the PR to do this.

vladimir-mencl-eresearch:
Now that the exception catching block is limited to float casting statements, does this look OK to merge?
Cheers,
Vlad

PR #87, opened 2021-09-01
Clickjacking protection

vladimir-mencl-eresearch:
Hi @zmousm ,

I hope you are doing well.

I've just noticed (from output of a security scanner against our DjNRO instance) that DjNRO is not issuing the X-Frame-Options header - which has become standard expectation of web applications these days.

And I found the middleware that would take care of that has been in the project `settings.py` file the whole time, just commented out.

This PR just uncomments it - and does a slight reorder to extend the protection also to flag pages.

Successfully tested in our TEST environment - reports a clean pass from the scanner, and no functionality broke.

Does that look OK to you to merge?

Cheers,
Vlad


zmousm:
Sure @vladimir-mencl-eresearch, I am not sure why we never enabled it.

PR #93, opened 2022-12-28
Differentiate preproduction/test sites in mapping

ghalse:
The existing mapping code displays preproduction/test sites on the home page, but does not mark them differently. This leads users to expect working eduroam at those sites.

This pull request does two things:
1) it updates the existing marker to a higher resolution one (the original SVG being under CC0) to make the map display better; and 
2) it introduces a new marker for preproduction/testing sites to make this clearer to end users.


ghalse:
Rebasedt to current master

vladimir-mencl-eresearch:
Hi @ghalse ,

Thanks, this looks great.

Two minor points:
(1) I see the code deals with the pin image as being 50x80, but the image is actually 51x80.  While the pre-prod image is (correctly sized as) 50x80.  Perhaps just an extra row of pixels that got accidentally included in `edupin.png` ?

(2) Would it be possible / reasonably easy to apply the same change also to the map displayed at `/manage` (for consistency) ?

I see the data served at `/manage/services/points/` does include the `stage` attribute, so hopefully might be just the matter of applying the same changes also to `djnro/static/js/management-map.js`.

Please let me know if this sounds reasonable to you.

Cheers,
Vlad

ghalse:
I've not actually looked at this for some time, so I'll need to work out how it might differ from what I'm currently using in our fork. But sure, sounds reasonable. I'm planning on trying to carve out some time to look at your most recent changes in the coming week with a view to rebasing our fork. So I'll add it to my to-do list for then.

PR #95, opened 2023-11-03
new: pass ATTRIB-INSTITUTION-TYPE to CAT API NEWINST call

vladimir-mencl-eresearch:
To exactly match  inst types as expected by CAT, define these in utils/cat_helper.py.

In edumanage.views.catenroll, add ATTRIB-INSTITUTION-TYPE with the correct value to params passed to CatEnroll.newinst.

As the params passed will go through deuglify, pass the param in notation expected by deuglify.

### DIFF
--- a/utils/cat_helper.py
+++ b/utils/cat_helper.py
@@ -3,6 +3,10 @@ from lxml import objectify
 from lxml.etree import XMLSyntaxError
 import re
 
+# Define names matching Institution ERType values exactly as expected by CAT API
+# https://github.com/GEANT/CAT/blob/master/core/IdP.php#L55-L57
+ERTYPE_CAT_API_NAMES = { 1: 'IdP', 2: 'SP', 3: 'IdPSP' }
+
 # http://code.activestate.com/recipes/135435-sort-a-string-using-numeric-order/
 def string_split_by_numbers(x):
     r = re.compile('(\d+)')
--- a/utils/cat_helper.py
+++ b/utils/cat_helper.py
@@ -3,10 +3,6 @@ from lxml import objectify
 from lxml.etree import XMLSyntaxError
 import re
 
-# Define names matching Institution ERType values exactly as expected by CAT API
-# https://github.com/GEANT/CAT/blob/master/core/IdP.php#L55-L57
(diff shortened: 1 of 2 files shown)


vladimir-mencl-eresearch:
Hi @zmousm , can you please have a look at this PR?

It fixes the issue discussed on Slack (DjNRO unable to create new organisations in CAT since August 2022) - and has been tested and works.

Please let me know if OK to merge.

Cheers,
Vlad

utils/cat_helper.py
```diff
@@ -3,6 +3,10 @@
 from lxml.etree import XMLSyntaxError
 import re
 
+# Define names matching Institution ERType values exactly as expected by CAT API
+# https://github.com/GEANT/CAT/blob/master/core/IdP.php#L55-L57
+ERTYPE_CAT_API_NAMES = { 1: 'IdP', 2: 'SP', 3: 'IdPSP' }
```
zmousm: We already have `edumanage.models.get_ertype_string` to translate a numeric institution type to string, which was required for eduroam DB (EDB) v2.

There is a minor inconsistency however between IdP+SP in EDB vs. IdPSP in CAT. But I think this does not warrant having a completely separate mapping for the same thing.

I will modify your patch to use the existing facility.

zmousm:
Hi @vladimir-mencl-eresearch. Indeed we somehow missed this change in CAT 2.1. This reminds me we would need to employ versioning for the CAT admin API, but anyway...

Please see the inline comment and the changes to your patch.

zmousm:
I can not push to your fork branch. Please ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Or see the branch `pr95` ([diff](https://github.com/REANNZ/djnro/compare/cat_inst_type...grnet:djnro:pr95)).

vladimir-mencl-eresearch:
Thanks for the quick response @zmousm !

Sorry, messed up permissions - I've merged your commits into this branch.

Happy to merge as it is now?

Cheers,
Vlad

PR #96, opened 2023-11-07
fix: edumanage/cat_enroll: fix add_option_to_params

vladimir-mencl-eresearch:
Add substitution missed in refactoring.


vladimir-mencl-eresearch:
Trivial fix, merging.

PR #97, opened 2023-11-09
Removal of Pebduroam from DJNRO

PaulLocksley:
Hi Team, 

I noticed recently that the Pebduroam link in the services menu is dead, with the state of pebble I don't see it coming back. 

Due to this I believe it should be removed from the template and install instructions. 

Please let me know if there is a better way to raise this concern/ request.


PaulLocksley:
I accidently added an extra commit to the branch not realising it would show up here when messing around. 
woops. 
So have closed request but still think this is something that should be done.

PR #98, opened 2023-11-09
new: edumanage/template: add link to GEANT Location codes

vladimir-mencl-eresearch:
On the Institution Edit page, add a link to wiki page maintained by GEANT listing the location codes that can be used for the Institution Type.

Make this text translatable.

Link to the page via Confluence tiny URL - which is equivalent to a page ID, but redirects to readable URL with page title.

### DIFF
--- a/djnro/templates/edumanage/institution_edit.html
+++ b/djnro/templates/edumanage/institution_edit.html
@@ -92,7 +92,7 @@
 				<div class="controls">
 					{{ form.venue_info }}
 					{% if form.venue_info.errors %} <div class="alert-danger"> {{ form.venue_info.errors|join_with_linebreaks }} </div>
-					{% endif %} <span class="help-block"> {{ form.venue_info.help_text }}</span>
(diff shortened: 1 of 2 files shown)


vladimir-mencl-eresearch:
Hi @zmousm , as per our Slack discussion.

The Confluence tiny URL sounds as the best option - it is [essentially an encoded page ID](https://confluence.atlassian.com/confkb/the-differences-between-various-url-formats-for-a-confluence-page-278692715.html), but it afterwards redirects to the "nicer" title-based URL.

I have not added this to the translations yet - I don't know the transiflex setup, can I leave that part with you?

Cheers,
Vlad

zmousm:
I will review this shortly

vladimir-mencl-eresearch:
Hi @zmousm , I've now realised the text should be both on Edit Institution and Edit Service  Location pages, so I've amended the commit accordingly.

Cheers,
Vlad

abtu:
This looks ok to me - richard@reannz

PR #100, opened 2024-04-03
Remove metions of dead pebduroam project from DJNRO template

PaulLocksley:
Hi Team, 

I noticed recently that the Pebduroam link in the services menu is dead, with the state of pebble I don't see it coming back.

Due to this I believe it should be removed from the template and install instructions.

Please let me know if there is a better way to raise this concern/ request.


vladimir-mencl-eresearch:
Thanks @PaulLocksley !

Given that Pebble Watch has been discontinued, the [company shut down](https://en.wikipedia.org/wiki/Pebble_(watch)), everything transferred to FitBit - and the links no longer work and never again will, removing them sounds the right thing to do.

@zmousm , anything you'd think needs to be added to the PR?  It looks good to me and ready to merge...

Cheers,
Vlad

vladimir-mencl-eresearch:
Thanks @PaulLocksley !

PR #103, opened 2024-08-20
Restrict validity of session and CSRF cookies

vladimir-mencl-eresearch:
By default, session cookies last for 2 weeks and CSRF cookies for 1 year:

https://docs.djangoproject.com/en/1.11/ref/settings/#session-cookie-age 
https://docs.djangoproject.com/en/1.11/ref/settings/#csrf-cookie-age

Restrict session and CSRF cookie lifetime to 1 hour (but let session auto-extend while there's activity)


James-REANNZ:
Looks good to me. Django's reasoning for the default makes *some* sense, but no browser is going to cache a page for an entire year.

PR #105, opened 2024-09-13
Fix timestamp handling in instxml and in data manipulation

vladimir-mencl-eresearch:
Fix pending issues identified in #92 ( [this comment](https://github.com/grnet/djnro/pull/92#issuecomment-2172242587) ) to avoid eDB ignoring updates to Institution or Realm data because `ts` has not changed:

(1) For an Institution, where `InstServer` and `ServiceLoc` objects track their own timestamps, return institution `ts` as maximum over `ts` of institution itself and its subordinate objects.

(2) In data manipulation of subordinate objects included in the export, update `Institution` `ts` whenever a subordinate object is added, updated or deleted.  
* This also covers *removing* an InstServer* or *ServiceLoc* which would not be covered by the `max` calculation.
* Achieve this via Django signals (these are also active for manipulations done via Django admin interface).
* Generally rely on `post_save` and `post_delete` signals.
* For `Contact` objects (linked via a `ManyToManyField`), use a `pre_delete` signal instead to still see the links.
* Guard `Institution.institutiondetails` navigation with `InstitutionDetails.DoesNotExist` (in case the details object does not exist).
* Suppress the signals from `parse_institution_xml` same as other update-related signals are suppressed (to avoid too high churn updating institution timestamp).

### DIFF
--- a/edumanage/management/commands/parse_institution_xml.py
+++ b/edumanage/management/commands/parse_institution_xml.py
@@ -26,7 +26,11 @@ from edumanage.models import *
 from edumanage.views import ourPoints
 from edumanage.signals import (
     disable_signals,
-    DUID_RECACHE_OURPOINTS, DUID_SAVE_SERVICELOC_LATLON_CACHE
+    DUID_RECACHE_OURPOINTS, DUID_SAVE_SERVICELOC_LATLON_CACHE,
+    DUID_INSTREALM_UPDATE_INST_TS,
+    DUID_SERVICELOC_UPDATE_INST_TS,
+    DUID_INSTSERVER_UPDATE_INST_TS,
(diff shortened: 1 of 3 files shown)


vladimir-mencl-eresearch:
I have tested this carefully and it solves the issues with CAT eDB updates.

@zmousm , @ghalse , any comments on this?  OK to merge?

ghalse:
I've had a quick look through it, and it seems to make sense. But I'm not going to get a chance to properly test it for some time :(

vladimir-mencl-eresearch:
Thanks @ghalse !

Has been now running in PROD at member.eduroam.net.nz for a week - with eDB ingesting the feed and timestamps being updated.

PR #108, opened 2025-01-10
Add support for RADIUS TLS-PSK

ghalse:
With the advent of [Blast-RADIUS](https://www.blastradius.fail/) there have been strong moves towards using RADIUS over TLS, and increased vendor support. The traditional mutual certificate auth method is hard to configure, but [RADIUS TLS-PSK](https://datatracker.ietf.org/doc/draft-ietf-radext-tls-psk/) is a promising alternative. It is already supported by both [FreeRADIUS](https://networkradius.com/doc/current/raddb/tls/tls-config_tls-common.html) and [radsecproxy](https://github.com/radsecproxy/radsecproxy/issues/112).

To allow institutions to configure RADSEC TLS-PKS, DjNRO needs to be extended to accommodate a new RADIUS type and to configure a TLS PSK identity and a TLS PSK key. This PR provides for that.

There are some assumptions in the implementation:

## TLS PSK will use RADIUS-TLS
In order to avoid extending the `proto` field from 12 characters, I've opted to use an identifier of `radius-psk` in `RADPROTOS`. This is technically ambiguous because there should be options for `radius-tls-psk` and `radius-dtls-psk` in RADPROTOS. However, DTLS doesn't seem widely implemented and is commented out in DjNRO's stock config. I think it's safe to assume this will be RADIUS-TLS rather than DTLS. Of failing that, that NROs will support one or the other.

## PSK identity for the local server
The local FLR server's PSK identity is not user-configurable. Instead, it is generated as a Network Access Identifier from the institution's UUID and a pre-defined realm configured in local_settings.py (new `NRO_TLSPSK_REALM` setting).

This creates an opaque token in the user portion, as contemplated by [4.2 of draft-ietf-radext-tls-psk](https://www.ietf.org/archive/id/draft-dekok-radext-tls-psk-01.html#section-4.2). An example identifier is:

> 7553a4b5-641a-4993-996e-ad36f169c834@flrs.eduroam.home.arpa

The identity is exposed in the user interface, and in the servdata YAML. It can also be constructed from the database if needs be.

## PSK for a remote server
The remote IdP's PSK identity can be configured and is simply validated as a Network Access Identifier.
 
## Shared pre-shared key
In theory, the IdP and SP connections can be configured completely independently. This means that a pair of servers can have a different RADIUS secret and/or a different PSK key for each direction.

However, DjNRO already assumes that both directions will share a common RADIUS secret. This also seems to an assumption made by many server admins. To avoid having to add two separate fields for PSK keys, this implementation extends that assumption by assuming that both directions will share a common TLS pre-shared key.

This isn't strictly in accordance with [draft-ietf-radext-tls-psk](https://datatracker.ietf.org/doc/draft-ietf-radext-tls-psk/), but it's a reasonable compromise between an ideal implementation and practical constraints. It will probably also reduce confusion.

Our [own implementation](https://github.com/tenet-ac-za/djnro/commit/464c13167f6bde25c784707497308a2f39bd6a85#diff-360fae78b91cfaeb27d3da5a92f9f6b9c34c49aa68119efc89ce57897a1ff93aR197-R208) of this extends things by having the initial PSK key pre-generate as a strong random string, rather than relying on the admin to supply one.


edumanage/forms.py
```diff
@@ -195,6 +197,38 @@ def clean_host(self):
...
+                    raise forms.ValidationError(_('This field is required for IdPs.'))
+
+    def clean_psk_key(self):
+        psk_key = self.cleaned_data['psk_key']
+        if not 'proto' in self.cleaned_data:
+                raise forms.ValidationError(_('The Protocol field is required to validate this field.'))
+        proto = self.cleaned_data['proto']
+        if proto in RADPROTOS.TLSPSK:
+            if psk_key:
+                if len(psk_key) > 16:
```
vladimir-mencl-eresearch: This would reject a key of exactly 16 octets - how about:
```suggestion
                if len(psk_key) >= 16:
```

edumanage/models.py
```diff
@@ -542,6 +543,8 @@ class InstServer(models.Model):
 
     secret = models.CharField(max_length=80)
     proto = models.CharField(max_length=12, choices=RADPROTOS, default=RADPROTOS.UDP)
+    psk_identity = models.CharField(max_length=64, null=True, blank=True, help_text=_("Network Access Identifier (user@realm)"))
```
vladimir-mencl-eresearch: I see draft-ietf-radext-tls-psk specifically asks that:
> Implementations MUST support PSK Identities of 128 octets
.... perhaps this should be
```suggestion
    psk_identity = models.CharField(max_length=128, null=True, blank=True, help_text=_("Network Access Identifier (user@realm)"))
```
(and migration changed too) ?

edumanage/migrations/0012_instserver_psk.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.29 on 2025-01-08 11:14
```
vladimir-mencl-eresearch: Given that #109 got merged in the meantime, perhaps the migration here should be renamed to `0013` ?

edumanage/context_processors.py
```diff
@@ -18,6 +18,7 @@ def country_code(context):
         'GOOGLE_MAPS_API_KEY': settings.GOOGLE_MAPS_API_KEY if hasattr(settings,"GOOGLE_MAPS_API_KEY") else None,
         'SHIB_AUTH_ENTITLEMENT': settings.SHIB_AUTH_ENTITLEMENT if hasattr(settings,"SHIB_AUTH_ENTITLEMENT") else None,
         'FEDERATION_DOC_URL': settings.FEDERATION_DOC_URL if hasattr(settings,"FEDERATION_DOC_URL") else None,
+        'OUR_TLSPSK_REALM': settings.NRO_TLSPSK_REALM if hasattr(settings,"NRO_TLSPSK_REALM") else 'localhost',
```
vladimir-mencl-eresearch: How about putting the default value for `NRO_TLSPSK_REALM` into `djnro/settings.py`, with site-specific value being configured in `djnro/local_settings.py` ?

The settings in DjNRO have been designed specifically for this - that local_settings.py can override values from settings.py.

This would avoid the need to have the default value hard-coded in 3 different places - and the code would be able to use `settings.NRO_TLSPSK_REALM` without having to check if it is defined, as `setting.py` comes with the code.

vladimir-mencl-eresearch (changes requested):
Thanks for the contribution, @ghalse .

Minor comments I've pointed out in the code - otherwise looks well designed.

Shouldn't be too much to address my concerns.

Cheers,
Vlad

edumanage/models.py
ghalse: The length was based on the NAI implementation, but actually this is probably better.

edumanage/forms.py
ghalse: Doh, that's a stupid oversight. Comes from never using < 20 chars to test :)

edumanage/migrations/0012_instserver_psk.py
ghalse: rebased to include #109 and fixed

edumanage/context_processors.py
ghalse: This one I'm in two minds about this. We wouldn't ever want anyone leaving this as `localhost` in production; that *should* be an error. So maybe removing the defaulting is better? Or maybe we need to make the default in settings.conf something obviously wrong in a reserved domain.

ghalse: I realise that a default must be set for people upgrading, so it's now deliberately invalid and uses an RFC2606 suffix.

djnro/local_settings.py.dist
```diff
@@ -152,6 +152,8 @@ NRO_PROV_SOCIAL_MEDIA_CONTACT = [
 
 # Helpdesk, used in base.html:
 NRO_DOMAIN_HELPDESK_DICT = {"name": _ld({'en': "Domain Helpdesk"}), 'email':'helpdesk@example.com', 'phone': '12324567890', 'uri': 'helpdesk.example.com'}
+# ream used to generate a TLS-PSK identity for service providers
+NRO_TLSPSK_REALM = "example.com"
```
Razorfang, line 156: How come the default values for `NRO_TLSPSK_REALM` is different in `settings.py` than it is here?

vladimir-mencl-eresearch: Hi @Razorfang , this is a change @ghalse did on my suggestion.

The value in settings.py is an intentionally invalid value that is used only when local_settings.py does not provide a value for `NRO_TLSPSK_REALM` and is there to prevent the code from breaking.

This is a value to be customised by local deployments.  Existing deployments that do not update local_settings.py will get the clearly invalid value.

As the two values server different purposes, it makes sense to me to have them different.

edumanage/views.py
```diff
@@ -2194,6 +2195,11 @@ def instxml(request, version):
                         server_type = EDB_SERVER_TYPES.TLS
                 except:
                     pass
+                try:
+                    if server.proto == RADPROTOS.TLSPSK:
+                        server_type = EDB_SERVER_TYPES.TLS
```
Razorfang, line 2200: Looking at the code here, what would happen if `server_type` was undefined as a result of `server.proto` not matching any of the above cases? Wouldn't the call to `instServerType.text = "%d" % (server_type)` fail?

ghalse: My PR doesn't introduce a new problem. However, I also think it's safe because of the check at lines [2184-2187](https://github.com/grnet/djnro/pull/108/files/70b950c407db4f5f746a9a7e0b1fc20d47a3e0a1#diff-7615b45d29866815d06ca6f6a10410abc92e1a84ddd59d14e8de267cc07bbfe3R2184-R2187). It'll simply become `EDB_SERVER_TYPES.UDP`.

djnro/local_settings.py.dist
ghalse: Yes, as @vladimir-mencl-eresearch suggests, they're intentionally different. The local_settings.py.dist file contains example values and uses the same example domain as the rest of the file (and fortunately that's already a documentation domain). The settings.py file uses an intentionally invalid value and uses a domain reserved for that purpose. Both are consistent with RFC2606.

edumanage/views.py
vladimir-mencl-eresearch: Thanks for the reply, @ghalse  - yes,  the agree lines 2184-2187 handle that well by setting the default value.

edumanage/forms.py
```diff
@@ -195,6 +197,38 @@ def clean_host(self):
...
 
+    def clean_psk_identity(self):
+        psk_identity = self.cleaned_data['psk_identity']
+        if not 'proto' in self.cleaned_data:
+                raise forms.ValidationError(_('The Protocol field is required to validate this field.'))
+        proto = self.cleaned_data['proto']
+        if not 'ertype' in self.cleaned_data:
+                raise forms.ValidationError(_('The Type field is required to validate this field.'))
+        ertype = self.cleaned_data['ertype']
+        if proto in RADPROTOS.TLSPSK:
```
vladimir-mencl-eresearch: Sorry @ghalse , I should have brought this up in my initial review - had it in my notes but missed it when submitting  the review.

This line would check for substring match - I think it should really be: 
```suggestion
        if proto == RADPROTOS.TLSPSK:
```

... as is e.g. in `edumanage/views.py` and `edumanage/management/commands/servdata.py`

I've seen the `proto in RADPROTOS.TLSPSK` pattern here (twice) and also in `djnro/templates/edumanage/server_details.html` - believe it should be `==` instead.

Cheers,
Vlad

ghalse: You're right. I guess this was me stylistically trying to match the existing `if server.ertype in ERTYPE_ROLES.IDP` pattern without thinking through the distinction. Fixed.

vladimir-mencl-eresearch (approved):

PR #110, opened 2025-01-29
fix: correctly export venue_info for InstDet and ServiceLoc

vladimir-mencl-eresearch:
Export was missed due to a typo - assign the value into the `.text` attribute of the XML element.


Razorfang:
Did this mess up the entire export or just the venue info?

vladimir-mencl-eresearch:
Just the venue info, producing empty `<inst_type/>` and `<location_type/>` - but only when a value is in the database, leaving it out otherwise.

After this change, the value would actually be rendered inside the element - so `<inst_type>2,7</inst_type>` and `<location_type>2,7</location_type>`

Razorfang:
I think it looks okay then.

PR #111, opened 2025-02-03
Update django-registration for django 1.11+

ghalse:
In the update to Django 1.11 (PR #56), django-registration was left behind. That hasn't caused any real problems except when you run `./manage.py cleanupregistration`. Then you get an error:

```
File "/opt/djnro/.venv/lib/python3.8/site-packages/registration/management/commands/cleanupregistration.py", line 10, in <module>
    from django.core.management.base import NoArgsCommand
ImportError: cannot import name 'NoArgsCommand' from 'django.core.management.base' (/opt/djnro/.venv/lib/python3.8/site-packages/django/core/management/base.py)
```

That is a result of a deprecation in Django 1.9 that went away in 1.11. It was fixed in django-registration 2.1.

Looking at the release notes for django-registration, there are no breaking changes that would affect DjNRO [all the way up to 2.5.2](https://django-registration.readthedocs.io/en/2.5.2/upgrade.html). None of the changed functions mentioned in the notes are used in DjNRO. So, pending a larger overhaul, it seems safe to bump this up to < 2.6. My testing hasn't revealed any problems with doing this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 Django>=1.11,<2.0
 argparse>=1.2.1
-django-registration==2.0.4
+django-registration>=2.1,<2.6
 django-tinymce>=2.3.0,<3.0a
 ipaddress==1.0.22
 social-auth-app-django>=3.1.0,<4.0


vladimir-mencl-eresearch:
Hi @ghalse , 

Thanks for this contribution - looks like I indeed missed updating django-registration in my earlier work on Django 1.11 upgrade.

My colleague @razorfang is now working on upgrading DjNRO to newer Django - so far has it working at Django 2.1.5 and is about to open a PR really soon now.

His work will also take care of `django-registration` upgrade, so this PR should not then be needed (and right now would clash with his work).

I'll keep you posted - but it should not take long before this further work is in.

Cheers,
Vlad

ghalse:
Sure, I see #112 which would supersede this. Happy to close it.

PR #114, opened 2025-02-20
Update project to work with Django 4.2.19

Razorfang:
Django 4.2 is a stable release that will be supported until April next year. These changes bring the project up to Django 4.2.19, which is currently the latest version of that release.


edumanage/urls.py
```diff
@@ -3,7 +3,7 @@
 from django.contrib.flatpages import views
 
 urlpatterns = [
-    re_path(r'^$', edumanage.views.index, name="index"),
```
vladimir-mencl-eresearch, line 6: Shouldn't `re_path` be also dropped from the imports (like it is in `djrno/urls.py` ?

requirements.txt
```diff
@@ -1,22 +1,18 @@
-Django==2.1.5
-argparse>=1.2.1
-ipaddress>=1.0.22
+# Django 4.2.19 is part of the Django 4.2 LTS and will reach end of life April 2026
+Django==4.2.19
```
vladimir-mencl-eresearch: Before the upgrade, `requirements.txt` was letting point releases of Django in - to allow any newer 1.11.x, it had: `Django>=1.11,<2.0`.
How about we accept newer 4.2.x with
```suggestion
Django>=4.2.19,<5.0
```

vladimir-mencl-eresearch (changes requested):
Hi @Razorfang ,

Solid piece of work - only two minor comments from me.  Thanks for sharing the detailed commit history, allowed me to see the individual changes and motivation for them.

Let's wait for others to chip in, but thumbs up from me.

Cheers,
Vlad

vladimir-mencl-eresearch:
Thanks for the work @Razorfang .

Looks good to me. @ghalse @PaulLocksley @zmousm , can you please have a look too?

requirements.txt
Razorfang: > Before the upgrade, `requirements.txt` was letting point releases of Django in - to allow any newer 1.11.x, it had: `Django>=1.11,<2.0`. How about we accept newer 4.2.x with

4.2.x is the LTS releases, so I don't think that adding `,<5.0` would be quite right. I think what we want is `Django>=4.2.19,<4.3`

edumanage/urls.py
Razorfang: It probably should. Will address tomorrow.

requirements.txt
vladimir-mencl-eresearch: Thanks for the feedback.

I went by what the previous rule was (`<2.0` for 1.11.x), but I'm not sure what the exact motivation behind that was.  From the announced Django versions, there won't be any 4.x past 4.2.x, so `<4.3` will work the same as `<5.0`.

I'm happy with either - just wanted to make it accept newer  4.2.x (as it is still a supported release and newer versions may come out).

`Django~=4.2.19` should work too - leaving specific syntax up to you.

edumanage/urls.py
Razorfang: Done

requirements.txt
Razorfang: Done

vladimir-mencl-eresearch (approved):
Thanks for the fix ups - now looks all good to me.

PaulLocksley:
Hey @vladimir-mencl-eresearch  and @Razorfang, I've been running this for a few days in our dev environment with a copy of our production database.
All the issues I found have now been resolved by working with @Razorfang and it now works better than ever.

I'm very happy with these changes.

vladimir-mencl-eresearch:
Thanks @PaulLocksley !

@ghalse , @zmousm , can I please get your view too?

ghalse (changes requested):
The [memcache example in local_settings.py](https://github.com/grnet/djnro/pull/114/files#diff-460d7c7ece6b02f5ee77f277e8df0199431b94bafbdcbd73a44e543681877568L112-L119) no longer works with django 4.2. Per [the docs](https://docs.djangoproject.com/en/4.2/topics/cache/#memcached) it changed in 3.2 and should now be `django.core.cache.backends.memcached.PyMemcacheCache` or `django.core.cache.backends.memcached.PyLibMCCache`. So at the very least, the local_settings.py.dist needs to be updated.

However, this might be symptomatic of a larger caching problem since when I do that, the `manage.py fetchkml` command fails:

```
(.venv) (DEV)eduroam@web-cpt-02:~/djnro-django-3-update-reannz$ ./manage.py fetch_kml
Fetching data from https://monitor.eduroam.org/kml/all.kml?gmaprnd=1740841105617...
Done fetching!
Updating cache
Traceback (most recent call last):
  File "/home/eduroam/djnro-django-3-update-reannz/./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/home/eduroam/djnro-django-3-update-reannz/edumanage/management/commands/fetch_kml.py", line 48, in handle
    self.refresh_cache(file)
  File "/home/eduroam/djnro-django-3-update-reannz/edumanage/management/commands/fetch_kml.py", line 79, in refresh_cache
    cache.set('points', bz2.compress(points), 60 * 3600 * 24)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/django/core/cache/backends/memcached.py", line 79, in set
    if not self._cache.set(key, value, self.get_backend_timeout(timeout)):
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/hash.py", line 344, in set
    return self._run_cmd("set", key, False, *args, **kwargs)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/hash.py", line 322, in _run_cmd
    return self._safely_run_func(client, func, default_val, *args, **kwargs)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/hash.py", line 211, in _safely_run_func
    result = func(*args, **kwargs)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/base.py", line 475, in set
    return self._store_cmd(b"set", {key: value}, expire, noreply, flags=flags)[key]
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/base.py", line 1247, in _store_cmd
    self._raise_errors(line, name)
  File "/home/eduroam/djnro-django-3-update-reannz/.venv/lib/python3.10/site-packages/pymemcache/client/base.py", line 1042, in _raise_errors
    raise MemcacheServerError(error)
pymemcache.exceptions.MemcacheServerError: b'object too large for cache'
```
That doesn't happen with `django.core.cache.backends.dummy.DummyCache`, nor did it happen with the old `django.core.cache.backends.memcached.MemcachedCache` (or maybe it just silently fails). But I get a similar error (pylibmc.TooBig) with `django.core.cache.backends.memcached.PyLibMCCache`.

I've not tested with Redis (now supported by django with this upgrade). But maybe that's a better thing to be suggesting in the config?

ghalse:
These changes mean that people must review their local_settings.py, and break backwards compatibility there. And the social-auth-app-django version specified requires python 3.10, meaning people who've worked around the old python 3.8 limitation with e.g deadsnakes will need to undo that. So I guess, ultimately, there needs to be a new version tagged and some release notes explaining what people need to do to upgrade. Not sure where that comes in under @Razorfang / @vladimir-mencl-eresearch 's plans for this update.

vladimir-mencl-eresearch:
Hi @ghalse ,

Many thanks for these insights - very helpful to get another perspective on testing.  Agree the release process should be managed, in particular with backwards compatibility with older local_settings.py breaking.

@Razorfang , can you please draft an upgrade guide that would in particular spell out changes required in local_settings.py ?

Under `docs/installation`, you can find existing upgrade guides - latest being `upgrading-from-1.1.md`.

Looking at the versions at https://github.com/grnet/djnro/tags, I think we should call what we merge `1.3.0` - and the new guide perhaps `upgrading-from-1.2.md`.

It would ideally also document changes merged in-between 1.2.1 and now - mostly coming from @ghalse and me.

I've just found a neat way to see all changes to `local_settings.py.dist` in context of their commits:
```
git show v1.2.1..django-3-update-reannz djnro/local_settings.py.dist
```
... might be useful to see what may require changes to locally deployed copies of `local_settings.py`.

Thanks for all the work on this!

Cheers,
Vlad

requirements.txt
```diff
@@ -1,22 +1,18 @@
...
 
-# Version 3.0 supports Django 1.11 through 2.1
-django-registration==3.0
+# These django packages are using versions that are known to work for Django 4.2
+django-registration==3.4
+django-tinymce==4.1.0
+social-auth-app-django==5.4.3
+social-auth-core==4.5.6
+django-sortedm2m==4.0.0
+django-widget-tweaks==1.5.0
```
vladimir-mencl-eresearch, line 10: Hi @Razorfang , just wondering whether to keep these exactly pinned or allow some floating along point releases like we did for Django.

Your thoughts?

Razorfang:
> Per [the docs](https://docs.djangoproject.com/en/4.2/topics/cache/#memcached) it changed in 3.2 and should now be `django.core.cache.backends.memcached.PyMemcacheCache` or `django.core.cache.backends.memcached.PyLibMCCache`. So at the very least, the local_settings.py.dist needs to be updated.

> However, this might be symptomatic of a larger caching problem since when I do that, the `manage.py fetchkml` command fails:

Thank you for reminding me about this, Guy. We had made changes to fix this for our own deployment (since we have our own copy of local_settings.py.dist that we use), but I didn't copy those changes back into the project. I'll update the cache line to PyMemcacheCache.

As for the out of memory issue, I was able to fix this in our deployment by creating a custom memcached.conf configuration file that increases the available memory. I can add this to the upgrade guide.

requirements.txt
Razorfang: I don't see the benefit in doing this for most packages. For Django it's fine because that's the top-level framework package and it will include security fixes more easily, and we know that they stick to semantic versioning very reliably. But for other packages, they're unlikely to receive significant revisions, and it's more stable to stick to versions that we know work.

PaulLocksley:
Following up on the cache discussion, I've deployed Redis on our instance and it works like a dream, it does still require 2 additional packages to be installed so it's not 0 config but probably easier than memcached.

Razorfang:
I have pushed some documentation updates.

vladimir-mencl-eresearch:
> I have pushed some documentation updates.

Thanks @Razorfang , this is looking great.

One comment on the upgrade guide.  You put together a detailed changelog - that is very helpful.

What might be still missing is a brief section on required changes to local_settings.py.  The information for that is already in the upgrade guide, but is spread over the changelog and not in one place.

Maybe adding an extra section on documenting the required changes, pointing back to the changelog to avoid repeating the same thing again?
The only setting that's NOT commented out by default is `NRO_TLSPSK_REALM`.
For that, the instructions could be "if supporting RADIUS-TLS-PSK connections, please set this variable to realm you want to use to generate a TLS-PSK identity for connections from eduroam SPs".
For memcached and for CAT, it could be "if you are using these options and have uncommented them from the default file, please update the values as documented in the changes section above", or something like that.

Otherwise, I think we're nearly there.

@ghalse ?

Cheers,
Vlad

djnro/settings.py
```diff
@@ -100,8 +104,10 @@
                 'edumanage.context_processors.country_code',
                 'edumanage.context_processors.cat_instances',
                 'edumanage.context_processors.manage_login_methods',
+                'edumanage.context_processors.nro_domain_url',
```
vladimir-mencl-eresearch: Sorry, only got to review this now.  It feels maybe too much to add a new context processor just to expose a single setting, given there is already a catch-all context processor, `country_code`.

Agree the name of that one is is not really fitting (must have been created first to expose just COUNTRY_CODE and later became the catch-all).

How about renaming `country_code` to something generic (? `djnro_settings` ?) and moving  `NRO_DOMAIN_MAIN_URL` in there?

Razorfang: Re-naming country code makes the most sense, since it's not actually related to just country codes any more. Will address now.

Razorfang: @vladimir-mencl-eresearch  done

vladimir-mencl-eresearch: Great, thanks for the quick action!

vladimir-mencl-eresearch:
Hi @ghalse , how does this look to you?  

@Razorfang has now also added upgrade documentation.

We have now rolled this out into our staging environment and haven't seen anything break.  The changes required to our deployment process (via Ansible)  match what has been covered in the upgrade documentation.

It looks ready to roll to me.  Please let me know if you see it ready to ship or if there's anything that needs correcting.

I would otherwise aim to merge this and call it a 1.3.0 release sometime soon.

Many thanks advance for getting back to me on this.

Cheers,
Vlad

vladimir-mencl-eresearch:
Hi @ghalse @zmousm ,

We've been running this code for a week with no issues. Unless there are any objections, I'll merge this PR tomorrow and tag a `v1.3.0` release.

Thanks for all the feedback with reviews @ghalse .

Cheers,
Vlad

PR #115, opened 2025-04-03
Fix config social for social login

vladimir-mencl-eresearch:
Hi @Razorfang ,

I found the sample config for social login needed some update to reflect changes in social_auth_core:
* Yahoo migrating from YahooOpenId backend to YahooOAuth2
* DropboxOAuth2 migrating from original to V2 API

Plus the logo URL we had for Docker no longer works, so I have downloaded a local image from their official logo bundle.

This is just a minor config change - can you please review?

Cheers,
Vlad


Razorfang:
Looks fine. I think we can merge this.

PR #112, opened 2025-02-03
Update Django framework to version 2.1.5

Razorfang:
This is the first of hopefully several updates to the version of Django used by the project.

These changes have been tested by REANNZ and seem to behave consistently with the master branch.

PR #116, opened 2025-05-16
Clean up from django_registration changes

vladimir-mencl-eresearch:
In the recent upgrades (#112, #114), Django registration was upgraded from 2.0.4 to 3.4  - and in between the versions, it moved from using a model to keeping activation code to cryptographic handling of activation codes without need for storage.

However, it did leave behind the `registration_registrationprofile`, not backed by any model anymore, but left with foreign key constraints against the `auth_user` table.

We ran into it when attempting to delete a user object created *before* the upgrade.

To support both existing deployments (left with an orphaned `registration_registrationprofile`), as well as new deployments where no such table exists, we saw as the best way forward adding a migration that deletes the table *if it exists*.

That brings both existing and new deployments to a consistent state where the `registration_registrationprofile` table, no longer needed, does not exist.

This PR also fixes uses of `user.registrationprofile` missed in #112, plus a minor glitch from the same PR that was passing `User` object instead of the `UserProfile` (and thus the activation page was missing the information on the institution the user requested access to).

@zmousm , @ghalse , would you please be able to review?


James-REANNZ:
Seems fairly simple, nothing jumps out at me as obviously wrong.

vladimir-mencl-eresearch:
Thanks @James-REANNZ !

PR #120, opened 2025-07-13
Show the value of is_social_active on the UserProfile admin page

Razorfang:
The variable _is_social_active_ is a good indication of whether a user account has been activated. This information is useful to have when looking through user accounts. This change adds the value to an additional column in the admin portal for convenience.

Screenshot 2025-07-14 at 11 15 19 AM

### DIFF
--- a/accounts/admin.py
+++ b/accounts/admin.py
@@ -4,6 +4,6 @@ from django.contrib.auth.admin import UserAdmin
 
 @admin.register(UserProfile)
 class UserPrAdmin(admin.ModelAdmin):
-    list_display = ('user', 'institution')
+    list_display = ('user', 'institution', 'is_social_active')
 
 admin.site.register(User, UserAdmin)
--- a/accounts/migrations/0001_initial.py
+++ b/accounts/migrations/0001_initial.py
@@ -43,7 +43,7 @@ class Migration(migrations.Migration):
             name='UserProfile',
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
-                ('is_social_active', models.BooleanField(default=False, verbose_name="Link to institution approved")),
+                ('is_social_active', models.BooleanField(default=False, verbose_name="Approved")),
             ],
             options={
                 'permissions': (('overview', 'Can see registered user and respective institutions'),),
--- a/accounts/migrations/0001_initial.py
+++ b/accounts/migrations/0001_initial.py
@@ -43,7 +43,7 @@ class Migration(migrations.Migration):
             name='UserProfile',
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
-                ('is_social_active', models.BooleanField(default=False)),
+                ('is_social_active', models.BooleanField(default=False, verbose_name="Link to institution approved")),
             ],
             options={
                 'permissions': (('overview', 'Can see registered user and respective institutions'),),
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -36,7 +36,7 @@ patch_username_maxlen(User._meta.get_field('username'))
 class UserProfile(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
     institution = models.ForeignKey(Institution, on_delete=models.CASCADE)
-    is_social_active = models.BooleanField(default=False)
+    is_social_active = models.BooleanField(default=False, verbose_name="Link to institution approved")
 
     class Meta:
         permissions = (
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -36,7 +36,7 @@ patch_username_maxlen(User._meta.get_field('username'))
 class UserProfile(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
     institution = models.ForeignKey(Institution, on_delete=models.CASCADE)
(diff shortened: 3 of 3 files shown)


vladimir-mencl-eresearch:
@ghalse @zmousm @paulhii your thoughts on how useful it would be?  Came from a request within our team to get an overview panel of account requests that haven't been processed yet.

ghalse:
It wouldn't be particularly useful to us; we don't have enough churn, and even if I were to notice it, it wouldn't change anything (other steps are required to address the issue). But I also don't think it's harmful, so if it's useful for others, then it's useful.

To me, what might make more sense (and I've not given any thought to implementation or practicability) is a way to see *and approve* the unapproved accounts from that page. Because if you've lost the email with the approval link, there are several steps you need to remember to replicate it.

paulhii:
This is useful for eduroam administrators/product owner like myself who only have full access on the admin web portal, i.e., no access to backend database to see the list of not is_social_active users. Therefore, at the moment, there is no easy way to see is_social_active unless i click into every User Profile. Its use is merely for reporting purposes. It's not going to help with determining if we have to enable social_active. Some users are intentionally left inactive. We don't have issue with activating users at the moment as we get email notifications (that's ticketed so they don't get lost) to action activations. Instead of an additional column, it might be useful to implement the filter that is currently available in the User admin interface so that I can get a report.

vladimir-mencl-eresearch:
Hi @ghalse , @paulhii , thanks for the feedback.

Paul, the difference is: `User.active` is a flag at the Django level signalling whether the account is enabled or disabled - an inactive account cannot log in at all.

`UserProfile.is_social_active` signals at DjNRO level whether the link between a User and an Institution has been approved (by an NRO admin).

@ghalse , I believe the way to approve an account without seeing the link (which now has a cryptographically derived token, not stored in a database) would be just to explicitly set the `is_social_active` flag.  @Razorfang , can you please confirm this is correct?

vladimir-mencl-eresearch:
On this note: would it help with understanding the meaning if the display name of the field was changed to something like `Link to Institution Approved` (alternative suggestions welcome) ?

@Razorfang , it should be doable with `short_description` in the Django Admin model: https://docs.djangoproject.com/en/4.2/ref/contrib/admin/

ghalse:
> @ghalse , I believe the way to approve an account without seeing the link (which now has a cryptographically derived token, not stored in a database) would be just to explicitly set the `is_social_active` flag. @Razorfang , can you please confirm this is correct?

Correct. But also then cleaning up in the Registration profiles.

vladimir-mencl-eresearch:
The registration profiles have disappeared in the upgrade to Django 4.2 - instead of storing the token in a database, the Django registrations module instead uses a cryptographically derived token (and the RegistrationProfile database model is gone).

So I believe it would be *only* setting the is_social_active flag.  Based on what I see in `accounts/views.py` in `activate` (lines 95-120), it is just setting is_social_active true (plus making sure `User.active` is true as well).  The only other thing that would be skipped is the email confirming the activation.

But in the user model, I believe setting is_social_active to true (assuming User.active is true) should give equivalent result.

paulhii:
> On this note: would it help with understanding the meaning if the display name of the field was changed to something like `Link to Institution Approved` (alternative suggestions welcome) ?
> 
> @Razorfang , it should be doable with `short_description` in the Django Admin model: https://docs.djangoproject.com/en/4.2/ref/contrib/admin/

@vladimir-mencl-eresearch thanks for clarifying the 2 flags. Yes, it will better understood if the field is named better or provide (i) infor button that explains its purpose. I confirm that we needed to turn on User active and is_social_active to complete activation manually when our email activation request broke. The 2 clicks could be 1 as they are not mutually exclusive for our use case of activating users.

Razorfang:
Thank you for your feedback, everyone. I have pushed changes that update the name of the field in the admin interface. This is done by modifying the `verbose_name` parameter of the `is_social_active` field.

This is what the change will look like when deployed.

This image shows the heading of the field changed to display `Link to institution approved`:

Screenshot 2025-08-05 at 12 22 37 PM

This image shows the same effect in the detailed view of an individual user profile:

Screenshot 2025-08-05 at 12 23 28 PM

vladimir-mencl-eresearch:
Thanks for the changes, @Razorfang !

Can someone (@ghalse , @paulhii @PaulLocksley ) please review the changes?

Given how much I was involved in shaping these changes, it should be someone other than me reviewing  these...

For the review, I see the questions as:
1. Whether it is overall useful for an NRO admin navigating the `/admin` interface - and looking for unapproved access requests.
2. Whether the name is fitting.  I could not come up with anything shorter and still meaningful.
3. Whether the change to an existing migration file is OK.  Adding the `verbose_name` parameter to the model would have triggered a new migration - which would have been a no-op, as the verbose name is not stored in the database model.  As there is no change to the underlying database model, I thought modifying the existing migration file is acceptable - and better than accumulating spurious migration files.

Many thanks in advance for reviewing this.

Cheers,
Vlad

James-REANNZ:
The actual code changes are pretty trivial, so my only feedback is that "Link to institution approved" is a bit word. "Approved" is clear enough in context. This has been double-checked with our resident eduroam admin and he agrees.

Razorfang:
Thanks for the review, @James-REANNZ . I'll push that change soon.

vladimir-mencl-eresearch:
Thanks for the review @James-REANNZ and for the quick action @Razorfang .

@James-REANNZ , can I please now get your seal of approval?

James-REANNZ:
Looks good to me.

PR #122, opened 2026-01-16
Update project to run on Django 5.2

Razorfang:
DjNRO currently runs on Django 4.2.

As per [the Django downloads page](https://www.djangoproject.com/download/#supported-versions), this version will reach end of extended support by April this year.

These changes update the project to use Django 5.2, which will receive extended support until April 2028.

This also fixes https://github.com/REANNZ/djnro/security/dependabot/18 by updating the version of social-auth-app-django to version 5.6. Updating Django to a later version was a requirement for this update.

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -9,55 +9,63 @@ from django.utils.translation import gettext_lazy as _
 from accounts.models import User
 from django.views.decorators.cache import never_cache
 from django import forms
-from django_registration.backends.activation.views import ActivationView
-from django_registration.exceptions import ActivationError
+from django.core import signing
 from accounts.models import UserProfile
 
 from edumanage.forms import UserProfileForm
 from edumanage.models import Institution
 
-logger = logging.getLogger(__name__)
+def get_user(username):
+    try:
+        user = User.objects.get(username__exact=username)
+    except User.DoesNotExist:
+        user = None
 
+    return user
+
+def render_activate_page(request, account):
+    return render(
+        request,
+        'registration/activate.html',
+        {
+            'account': account,
+            'expiration_days': settings.ACCOUNT_ACTIVATION_DAYS
+        }
+    )
 
 @never_cache
 def activate(request, activation_key):
     account = None
+    try:
+        username = signing.loads(
+            activation_key,
+            salt=settings.REGISTRATION_SALT,
+        )
+    except signing.BadSignature:
+        return render_activate_page(request, account)
+
     if request.method == "GET":
-        # Normalize before trying anything with it.
-        activation_view = ActivationView()
-        try:
-            username = activation_view.validate_key(activation_key=activation_key)
-        except ActivationError:
-            return render(
-                request,
-                'registration/activate.html',
-                {
-                    'account': account,
-                    'expiration_days': settings.ACCOUNT_ACTIVATION_DAYS
-                }
-            )
-        try:
-            # NOTE: This fix works for now but isn't that clean.
-            # This is required because when a new user is created with google-oauth2,
-            # they are marked with is_active = True, which will cause the call to
-            # get_user to fail because it only works when is_active = False.
-            temp_user = User.objects.get(username__exact=username)
-            if temp_user.is_active:
-                temp_user.is_active = False
-                temp_user.save()
-            user_profile = activation_view.get_user(username).userprofile
-        except ActivationError:
-            return render(
-                request,
-                'registration/activate.html',
-                {
-                    'account': account,
-                    'expiration_days': settings.ACCOUNT_ACTIVATION_DAYS
-                }
-            )
-        form = UserProfileForm(instance=user_profile)
+
+        user = get_user(username)
+        if user is None:
+            return render_activate_page(request, account)
+
+        # NOTE: This fix works for now but isn't that clean.
+        # This is required because when a new user is created with google-oauth2,
+        # they are marked with is_active = True, which will cause the call to
+        # get_user to fail because it only works when is_active = False.
+        if user.is_active:
+            user.is_active = False
+            user.save()
+
+        # Check if the user is already activated, which we do by checking the `is_social_active` value
+        user_profile = user.userprofile
+        if user_profile.is_social_active:
+            return render_activate_page(request, account)
+
+        form = UserProfileForm(instance=user.userprofile)
         form.fields['user'] = forms.ModelChoiceField(
-            queryset=User.objects.filter(pk=user_profile.user.pk), empty_label=None
+            queryset=User.objects.filter(pk=user.pk), empty_label=None
         )
         form.fields['institution'] = forms.ModelChoiceField(
             queryset=Institution.objects.all(), empty_label=None
@@ -73,7 +81,6 @@ def activate(request, activation_key):
         )
     if request.method == "POST":
         request_data = request.POST.copy()
-        activation_view = ActivationView()
         try:
             user = User.objects.get(pk=request_data['user'])
             up = user.userprofile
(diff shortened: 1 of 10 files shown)


djnro/templates/base.html
```diff
@@ -70,9 +70,14 @@
...
-                <li><a href="{% url 'logout' %}">Logout</a></li>
-                {% endif %}
+                <li>
+                  {% if request.user.is_authenticated %}
+                  <form method="POST" action="{% url 'logout' %}">
+                    {% csrf_token %}
+                    <button type="submit" class="btn btn-link">Logout</button>
+                  </form>
+                  {% endif %}
+                </li>
```
vladimir-mencl-eresearch, line 79: This looks like it will render an empty `<li></li>` when no user is logged in.  Shouldn't the `<li>` be inside the conditional block?

accounts/views.py
```diff
@@ -9,55 +9,63 @@
...
+        # they are marked with is_active = True, which will cause the call to
+        # get_user to fail because it only works when is_active = False.
+        if user.is_active:
+            user.is_active = False
+            user.save()
+
+        # Check if the user is already activated, which we do by checking the `is_social_active` value
+        user_profile = user.userprofile
+        if user_profile.is_social_active:
+            return render_activate_page(request, account)
```
vladimir-mencl-eresearch, line 64: When comparing the original code with this, I see the original code rendered the activation page when handling `ActivationError`, while here it is rendered when `user_profile.is_social_active` evaluates to `True`.

Just to confirm: is that correct / what was intended?

accounts/views.py
```diff
@@ -9,55 +9,63 @@
...
+    )
 
 @never_cache
 def activate(request, activation_key):
     account = None
+    try:
+        username = signing.loads(
+            activation_key,
+            salt=settings.REGISTRATION_SALT,
+        )
```
vladimir-mencl-eresearch, line 43: Looking into `django/core/signing.py`, I don't think you need to pass `salt` to `loads()`.  It uses it to initialise the `TimestampSigner` object, but is not really used in the signature checking.

And passing it like this would I believe trigger a `KeyError` if `REGISTRATION_SALT` was missing from `local_settings.py` (haven't checked, assume so).

Dropping it from the arguments here would reduce the extra dependency.

vladimir-mencl-eresearch (changes requested):
Hi @Razorfang ,

Thanks, solid piece of work - and just a few minor points / clarifications.

Well done!

Cheers,
Vlad

djnro/templates/base.html
Razorfang: Since this is for a logout button, isn't that consistent with the old behaviour? I didn't add or remove any additional `<li>` tags or conditions, so the behaviour should be the same. I also checked this on our production host that doesn't have my changes, and the logout button only shows up when you're logged in.

accounts/views.py
Razorfang: The old code would raise an `ActivationError` on these conditions:

- The user account does not exist. We cover this case by the new check on line 50.
- The user accout is already marked as active with the `is_active` field. This is always set to False at this point in the code because of the workaround for new google accounts always being marked active (see comment). We cover checking for active accounts on line 63 now.

This means we are covering both cases. The changes are intentional.

accounts/views.py
Razorfang: I will test the change and report back.

djnro/templates/base.html
vladimir-mencl-eresearch: Hi @Razorfang ,

In the original code, the `<li>` + `</li>` tags were *inside* the `{% if }`, now they're  *outside*.

That means that when the condition is false (used not logged in), the code still generates empty `<li></li>` (with whitespace in-between).

This gets silently ignored by the browser, but shouldn't be there - and depending on CSS  styles applied to `<li>`, it could also get some visual appearance.

You are correct no Logout button is rendered when it should not, but the HTML has extra `<li> </li>`.

accounts/views.py
vladimir-mencl-eresearch: Thanks for confirming!

djnro/templates/base.html
Razorfang: Okay, you're right about the tag and condition being swapped. I missed that part. I'll rearrange them to the original order and re-test.

accounts/views.py
Razorfang: So I've done some tests where we remove the salt value from the `loads` call, and it raises a `BadSignature` exception. The docs say "raises BadSignature if signature fails." That makes sense to me, since whenever the activation key is created by the call to get_activation_key, it's actually calling this under the hood (see [the docs here](https://django-registration.readthedocs.io/en/stable/_modules/django_registration/backends/activation/views.html#RegistrationView.get_activation_key))

```
signing.dumps(obj=user.get_username(), salt=REGISTRATION_SALT)
```

So because the key is created using a salt value with `dumps`, we need the same salt value to get the username back. So we *do* have to have this new code.

I understand this is frustrating because it requires us to include the registration salt in the new settings. But we could work around this by replacing the direct `settings.REGISTRATION_SALT` usage with a function called `get_registration_salt()` that returns the default value if the setting doesn't exist.

djnro/templates/base.html
Razorfang: Check the latest commit

vladimir-mencl-eresearch: Thanks, resolved now!

accounts/views.py
vladimir-mencl-eresearch: Thanks for the digging.

Sorry, I got it wrong: the `Signer.unsign` method indeed checks the signature by comparing a new signature with the one provided - and the new signature is constructed in `Signer.signature` with the salt.

The `loads` method uses `"django.core.signing"` as default value for `salt` - which is however different from the default salt value used by `django-registration`, `"registration"` (as per the [activation workflow documentation](https://django-registration.readthedocs.io/en/stable/activation-workflow.html#security-considerations)).

I assume just removing `REGISTRATION_SALT` from `local_settings.php` (to mimick a site upgraded to this code version without updating `local_settings.php`) would cause a KeyError in this code.

I see the following ways forward:
1. Explicitly check for missing `settings.REGISTRATION_SALT` and use `"registration"` value instead.
2. Include a default value of `REGISTRATION_SALT` in `djnro/settings.py` (to be overridden in `local_settings.py`).
3. Make the setting required in `local_settings.py`.

Thoughts?

Given that the real security is through `settings.SECRET_KEY`, I'm leaning towards 2 for simplest code and simplest upgrade procedure.

Razorfang: Approaches 1 and 2 accomplish the same thing (providing a default) but in different ways, but 2 is a bit cleaner and makes users more aware of what's changed. I agree with going for approach 2. I'll make the change now.

vladimir-mencl-eresearch: Thanks, agree on the evaluation.

Just saw the change come in  - sorry, I meant something slightly different:
* add a fallback value into `settings.py`
* but still encourage users to override it in `local_settings.py`

So I expected the setting in the `local_settings.py.dist` to stay.

This way:
* if someone runs with an outdated `local_settings.py` with no `REGISTRATION_SALT` setting, the value from `settings.py` would get used.
* otherwise, if `local_settings.py` has `REGISTRATION_SALT`, that value would get used.

For style, I'd suggest the one in `settings.py` looks like a real fall-back value - e.g.:
```
REGISTRATION_SALT = "django-registration"
```
while in `local_settings.py.dist`, it should clear instruct to customise the value:
```
REGISTRATION_SALT = "customise-this-value"
```

And then the documentation in `install.md` should state the above: that we strongly recommend adding a value into `local_settings.py`, and that a default value would be used otherwise.   Oh, and it might include a recipe for constructing a value - e.g., `openssl rand -base64 18` would do.

docs/installation/install.md
```diff
@@ -529,4 +529,14 @@ To fix this, edit the memcached configuration file (usually located at `/etc/mem
...
 
+## Account Registration Salt (DjNRO Version 1.3.1 and Later)
 
+DjNRO version 1.3.1 updates the project from running on Django 4.2 to Django 5.2. This required updating the package `django-registration` from version 3.4 to version 5.2.1.
+
+However, a consequence of this change is that the account registration and activation flow changed between package versions. To make DjNRO work with this change, a new setting has been added called `REGISTRATION_SALT`.
+
+This setting is used during the creation of activation keys for new users. For more information about this setting, refer to [the django-registration documentarion](https://django-registration.readthedocs.io/en/stable/activation-workflow.html#security-considerations).
+
+This setting is defined in the file `settings.py`. If you wish to overwrite this value, you can redefine the setting in `local_settings.py`. Changing this setting is not *strictly* necessary, as per the django-registration documentation. However, we **STRONGLY** recommend changing the default value.
```
vladimir-mencl-eresearch:
```suggestion
While `settings.py` provides a fallback value for this setting  and changing it is not *strictly* necessary, we *strongly* recommend overriding the value in `local_settings.py` to a locally generated value.  The current version of the `local_settings.py.dist` template includes a place-holder for this value.
```

Razorfang:
@vladimir-mencl-eresearch  Sorry for the repeated commits regarding the registration salt. I believe I've addressed all your comments now.

vladimir-mencl-eresearch (approved):
All of my comments have now been addressed - thanks @Razorfang !

vladimir-mencl-eresearch:
Thanks @Razorfang for this work upgrading the code base to Django 5.2.

I've now reviewed it and I'm happy with what the code looks like.

@ghalse , @zmousm , could you please cast your eyes over this PR as well?

It is just a routine framework upgrade with minor changes, so I hope to be able to merge it within a week, but if I can get another pair of eyes, that would be much appreciated.

Cheers,
Vlad

PR #125, opened 2026-03-25
new: update deps, auto-accept point-releases

vladimir-mencl-eresearch:
Update to requests 2.33.0 (for CVE-2026-25645)

Update also django-widget-tweaks and lxml

Change version specifiers to `~=` (Compatible Release), automatically accepting point releases on the same minor release.


Razorfang:
Is there a reason the `Mako` package has its capitalisation changed?

vladimir-mencl-eresearch:
> Is there a reason the `Mako` package has its capitalisation changed?

Good catch!  I had thought it had wrong capitalization because it resolved to lower-case `mako`, but I see official documentation also says `Mako` - and `Django` is also listed with capital D but resolves to `django`.

Reverted to `mako`.

Razorfang:
That resolves everything for me. Looks good. Feel free to merge.

PR #129, opened 2026-05-20
new: indent (pretty-print) generated XML

### LINKED ISSUE #128: Provide an option to 'prettify' the XML export (for debugging purposes)
At the moment, institution.xml does not contain *any* newlines (so it appears as one big single line). For small NROs that's not a problem, but when attempting to debug issues with larger NROs, it would be nice to spit out the institution.xml in a 'prettified' format (every element on its own line with nesting indents). That way it would be easier to find specific issues when running the XML through the eduroam[.]org validator (although the size will be bigger because of the added whitespace). 👍

vladimir-mencl-eresearch:
Include also a newline when pre-pending the XML declaration.

Fixes #128

### DIFF
--- a/edumanage/views.py
+++ b/edumanage/views.py
@@ -2743,9 +2743,11 @@ def to_xml(ele, encoding="UTF-8"):
     Convert and return the XML for an *ele*
     (:class:`~xml.etree.ElementTree.Element`)
     with specified *encoding*.'''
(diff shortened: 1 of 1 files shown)


Razorfang:
Pretty simple change, I don't see anything that could go obviously wrong here. Looks good to me.

PR #130, opened 2026-05-20
Fix two issues in Django model

vladimir-mencl-eresearch:
(1) Fix choices for `Realm.country` - to avoid Django 5.2+ displaying a grouped list where language version of country name become choices (and country code gets replaced with language code)

(2) Stop Django from generating spurious migrations for accounts.User model.

Details in commit messages.


Razorfang:
The changes seem to be implemented in a way that won't cause spurious migrations, so this looks good to me.

PR #131, opened 2026-07-24
Use FEDERATION_NAME in management.html template

vladimir-mencl-eresearch:
We already have FEDERATION_NAME context parameter.

Use it in more places on the /management page, instead of literal "AAI Federation".

Adjust localisation file accordingly.

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -1044,7 +1044,7 @@ msgstr ""
 
 #: djnro/templates/front/management.html:14
 msgid ""
-"Administrators can log in using either Social Networks or the AAI Federation:"
+"Administrators can log in using either Social Networks or the %(FEDERATION_NAME)s:"
 msgstr ""
 
 #: djnro/templates/front/management.html:17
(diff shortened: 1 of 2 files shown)


vladimir-mencl-eresearch:
@Razorfang @ghalse could I please ask you for a review?

locale/en/LC_MESSAGES/django.po
```diff
@@ -1044,7 +1044,7 @@ msgstr ""
 
 #: djnro/templates/front/management.html:14
 msgid ""
-"Administrators can log in using either Social Networks or the AAI Federation:"
+"Administrators can log in using either Social Networks or the %(FEDERATION_NAME)s:"
```
Razorfang, line 1047: Should there be an `s` at the end here? What if the value of `FEDERATION_NAME` doesn't suit well to this format?

vladimir-mencl-eresearch: The `s` indicates string format in this syntax (as opposed to `d` for numeric).

Razorfang: Ah, my bad. In that case everything seems fine to me.

PR #132, opened 2026-09-16
ANSI_QUOTES aren't supported by default

theseal:
Causing migration error like:
MySQLdb.ProgrammingError: (1064, 'You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \'"registration_registrationprofile"\' at line 1')


vladimir-mencl-eresearch:
Thanks for the contribution @theseal .  Makes this custom migration DB vendor-independent.  Thanks for making this work also for MySQL.

PR #133, opened 2026-09-18
Easier overview of users

theseal:
Extend the Users view to list more useful attributes:
 * date_joined
 * last_login
 * is_social_active (Approved)
 * profile_link (with a clickable link)

Added filter to filter user by
 * Approved
 * Has Profile

Think this can be very useful when managing the user database and doing housekeeping


vladimir-mencl-eresearch:
Hi @theseal ,

Thanks, looks like a very useful addition.

I'm  asking for feedback from our team, but so far looks as a very neat improvement.

Cheers,
Vlad

Issue #118, opened 2025-07-02
A Users section to manage the institution users.


paulhii:
Currently, the Users of the institution is not managed by the institution because the institution has no visibility into who has access to DjNRO Admintool. NRO has to manually manage the Users. A Users section can be made available so users can see users (activated/pending activation) in their own institution who has access. 

It will be nice if the users will also be able to approve/activate their own users to access Admintool. The institution will have full autonomy to add/remove users who can manage their institution's eduroam deployment information. Today, an NRO like AARNet will activate users after we get confirmation from existing User of the institution.

The users should also be able to delete their own users especially those who are no longer part of their institution. Most of the time, NRO does not know if a User is no longer at the institution. As an NRO, this is a big issue as we are collecting personal identifiable information (e.g. email address) which we want institutions to be able to manage and keep up to date.

If institution cannot manage (activate/delete), then there needs to be a facility/information for them to request deletion of their users.

PR #134, opened 2026-09-18
Add a Admins view for Institution admins

theseal:
Add a separate view/page so that the institution admins can see other administrators from that org.

Optional (default off) feature to allow an admin to remove other admins from the same institution.


theseal:
Semi-related to #118

theseal:
I don't think it would be wise to do in this PR but I would also like to investigate on changing the approval flow from NRO admins to Institution admins. (As a feature flag of course).

So if you are discussing this PR have that in mind.

PR #135, opened 2026-09-18
Handle databases created before MariaDB 10.7+

theseal:
Django 5.0+ uses a native `uuid` column for UUIDField on MariaDB 10.7+ (has_native_uuid_field), instead of the char(32) hex representation used everywhere else. Databases whose edumanage_institution/edumanage_serviceloc tables were created before the server was upgraded to MariaDB 10.7+ still have the old char(32) columns, which are too short for the value Django now sends (a 36-character dashed UUID), causing "Data too long" errors.


vladimir-mencl-eresearch:
Thanks for the fix!  Looks all reasonable - while PostgreSQL also supports a native UUID column type, it's only MariaDB where this issue bites because UUID got introduced later.

PR #136, opened 2026-09-18
NRO administration improvements

theseal:
* Added link from /manage to NRO Administration. Link only showed when signed in as staff.
* Possibility to protect /admin with Shibboleth in order to not show the default django login screen if a administrator access /admin without a django session already


vladimir-mencl-eresearch:
Hi @theseal ,

Thanks for the contribution!  I like the idea in general, but I'm reading the changes carefully to avoid being caught out by unintended consequences.

Deployments can quite vary in the how they handle user authentication - but I assume most would support federated (Shibboleth) login alongside other authentication providers (like Google).

That's where the tricky bit comes:
* having this change enabled *with the Shibboleth protection of /admin* would block access for non-Shibboleth  users
* having this change enabled *without the Shibboleth protection of /admin* would create a security hole: anyone could spoof the Shib headers.  The Shib module normally protects against such spoofing - but only if the page is in space protected by Shibboleth.

One could also consider Lazy Shibboleth sessions - but that would kill the purpose here, as you intend to reuse the session without explicit login.

So I find the `ShibAdminAutoLoginMiddleware` rather tricky (and dangerous) to use.  There should at the very least be a configurable setting (in local_settings.py) to enable it - and the default should be Off.  Only deployments that are Shibboleth-auth-only and that protect `/admin` would turn it on, after confirming they meet all pre-requisites.

What do you think?


I like the other part here, the  NRO admin link.  And that could also be an answer to the problem that the first part of this PR aims to solve: if the user authenticates to `/manage` first (with whatever authentication that uses), they'd be "one click away" from the admin interface.

Two thoughts here:
1. Would it be possible to make the hostname in the NRO Admin link configurable (instead of relative `/admin`)?   We expose the admin interface on a different hostname (`admin.<base hostname>`) - and other NROs might be doing something similar (if they wanted different access policies for where  the URL can be reached from).
2. Would it make sense to make the link/button visually blend into the existing interface - or does it intentionally look different?


Trying to look at it from perspective of various deployments.

Please let me know what you think.

Cheers,
Vlad

theseal:
> Thanks for the contribution! I like the idea in general, but I'm reading the changes carefully to avoid being caught out by unintended consequences.
> 
> Deployments can quite vary in the how they handle user authentication - but I assume most would support federated (Shibboleth) login alongside other authentication providers (like Google).
> 
> That's where the tricky bit comes:
> 
>     * having this change enabled _with the Shibboleth protection of /admin_ would block access for non-Shibboleth  users
> 
>     * having this change enabled _without the Shibboleth protection of /admin_ would create a security hole: anyone could spoof the Shib headers.  The Shib module normally protects against such spoofing - but only if the page is in space protected by Shibboleth.
> 
> 
> One could also consider Lazy Shibboleth sessions - but that would kill the purpose here, as you intend to reuse the session without explicit login.
> 
> So I find the `ShibAdminAutoLoginMiddleware` rather tricky (and dangerous) to use. There should at the very least be a configurable setting (in local_settings.py) to enable it - and the default should be Off. Only deployments that are Shibboleth-auth-only and that protect `/admin` would turn it on, after confirming they meet all pre-requisites.
> 
> What do you think?

You are right. I did focus too much on Shibboleth and not really acknowledge the other methods exists or the security implication that shib headers could introduce without a proper Shibboleth in front.

I will push an update to address this which will force all users through the "regular DJNRO" login flow instead. I think that will be secure enough 🙂 

(The only method we allow is Shibboleth)

> I like the other part here, the NRO admin link. And that could also be an answer to the problem that the first part of this PR aims to solve: if the user authenticates to `/manage` first (with whatever authentication that uses), they'd be "one click away" from the admin interface.
> 
> Two thoughts here:
> 
>     1. Would it be possible to make the hostname in the NRO Admin link configurable (instead of relative `/admin`)?   We expose the admin interface on a different hostname (`admin.<base hostname>`) - and other NROs might be doing something similar (if they wanted different access policies for where  the URL can be reached from).

Yes. Pushed!

>     2. Would it make sense to make the link/button visually blend into the existing interface - or does it intentionally look different?

It's intentional different to indicate that this is not part of the regular Institution management.

Don't have strong opinion about this and could go either way. Want me to change it?

djangobackends/admin_login_middleware.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if (
+            request.path.startswith('/admin/') and
+            not request.user.is_authenticated
+        ):
+            return redirect(
+                '%s?next=%s' % (reverse('manage_login_front'), request.path)
```
vladimir-mencl-eresearch: To be consistent with how this is processed in edumanage.views, we should use `REDIRECT_FIELD_NAME` from `django.contrib.auth` instead of hardcoded `'next'`.
```suggestion
                '%s?%s=%s' % (reverse('manage_login_front'), REDIRECT_FIELD_NAME, request.path)
```

djangobackends/admin_login_middleware.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*- vim:encoding=utf-8:
+# vim: tabstop=4:shiftwidth=4:softtabstop=4:expandtab
+
+from django.shortcuts import redirect
+from django.urls import reverse
+
```
vladimir-mencl-eresearch:
```suggestion
from django.contrib.auth import REDIRECT_FIELD_NAME

```
for use in the other suggested change

vladimir-mencl-eresearch (changes requested):
Thanks for these changes.  The configurable NRO_ADMIN_URL is much appreciated - exactly fits our needs.

Just a nit here about "next" being used as a string here but referring to `REDIRECT_FIELD_NAME` in `edumanage/views.py`.

I'll reply separately on the middleware part.

edumanage/views.py
```diff
@@ -1562,10 +1563,14 @@ def user_login(request):
...
                 login(request, user)
-                return HttpResponseRedirect(
-                    request.GET.get(REDIRECT_FIELD_NAME,
-                                    default=reverse('manage'))
-                )
+                redirect_to = request.GET.get(REDIRECT_FIELD_NAME)
+                if not redirect_to or not url_has_allowed_host_and_scheme(
+                    url=redirect_to,
+                    allowed_hosts={request.get_host()},
+                    require_https=request.is_secure(),
```
vladimir-mencl-eresearch: For sites that go by multiple names, include all their names as listed in `settings.ALLOWED_HOSTS`.

For determining whether HTTPS is required, `settings.SESSION_COOKIE_SECURE` may be a better proxy than just whether the current request was over HTTPS.  (Though these days, any site should be using HTTPS only *and marking cookies as secure*).

```suggestion
                    allowed_hosts=settings.ALLOWED_HOSTS,
                    require_https=settings.SESSION_COOKIE_SECURE,
```

vladimir-mencl-eresearch:
Thanks for the patience with me @theseal .

I'm wondering about what the actual need is for specialised middleware for the Admin login.

I'm a bit puzzled by the bit you put into the documentation:
> By default, `/admin/` uses Django's own username/password login form, even if you sign in to `/manage/` via Shibboleth SSO, a local account, or a social backend.

Just to confirm, the redirect of an authenticated user from the manage pages to `/admin/` already works for you (that the user ends up logged in there), and you are adding this middleware as a shortcut to trigger the login also when the user hits `/admin/` directly?

This might still not be a fit for all deployments: even when they use external SSO for Institution administrators, there might be a separate local login for the NRO admin (even if used just as a break-glass).

Can you make this functionality configurable (as an opt-in, with default being status quo) ?

Cheers,
Vlad

